Issue #91, opened 2015-03-14
Implement sorting filters


oparoz:
The default is to sort alphabetically, but it should be possible for the user to sort his collection by date and to reverse that order
#### Requirements
- [x] Sorting interface
- [x] Sorting code. Pull the data from the cache, send it to the client and process it.
- [x] Fix https://github.com/interfasys/galleryplus/issues/56
#### Open questions
- Should we have a dropdown with 4 possibilities? 2 buttons? Where?
- Which date to use? Currently only the uploading date is available, but if meta data support is introduced then it would be possible to sort by date-the-picture-was-taken
- Should we we have more filters? Sort by brightness, colour, size, etc.

Designers: @jancborchardt 
Interested parties: @deMattin @Riechao @noooz @seuf @FredrikLindseth @Aesculapius @simonclausen

deMattin:
For sorting I only see the filename, filedate and date picture is taken.
This could be a simple pulldown named "sorting" with this 6 options (every option and reverse).

The activated sorting should remain active as long as gallery is active and has to be reset to default if you restart gallery. I don't think, a user customizing of this "default" is neccessary.

With more meta data support it could be possible to **filter** the output.
I don't see the sorting in this context as a useful task.

But there are many questions remaining if you implement filters.
You have to get a list of existing (selected) meta-data in the current folder to set a filter only on existing values. How could such a "list" look like?

jancborchardt:
At first we don’t really need an interface for it. Sorting should be by date taken by default. Alphabetic sorting makes 0 sense for photos.

After that, we can still judge if we really need other options.

deMattin:
Alphabetic sorting does make sense!

Cameras save their picture names chronological in alphabetical order and if you have a collection of (renamed) photos it is normally also sorted alphabetically.
You can not really rely on the meta data "date taken" on cameras. They often have wrong dates because of previously empty batteries and not have setting correct date after such a "reset".

And if you rename, you have influence on sorting (thematic prefix).
Editing and correcting meta data is very often not done by many users.

So I definitly vote for alphabetical order as default as it is now!

Koire:
@deMattin There are 2 problems I've found with this:
1.) The default of the gallery is to sort A-Z (smallest to largest -- reverse chronological order), and I have 4 or 5 years worth of photos from various smartphones, viewing my more recent photos is a hassle.
[...]

PR #110, opened 2015-04-07
First shot at a sorting interface

### LINKED ISSUE #90: [enhancement request ] Reverse order
As I have many pictures in my account I miss a feature to revert the order of the previews.

In a folder with  many folders it shows (and generates) the oldest folders/pictures first because of A-Z order.
But often (in my case "mostly") I want to see the **newest** pictures and folder first (Z-A).

So it would be a nice feature, to have the opportunity to revert display, transfer and also generation of the previews.

A simple switching caret symbol like in the files app would be the only modification in GUI and I think reverting the complete routine from A-Z to Z-A could not be much enhancement in code because there is a sorting of this already implemented (at a central point?) as it is now already alphabetical.

Would this be as simple as I think and do you like the idea?

**Edit:**
Is this the part of "gallery.js" to be modified?

``` javascript
// fill the albums from Gallery.images
Gallery.fillAlbums = function () {
    var sortFunction = function (a, b) {
        return a.path.toLowerCase().localeCompare(b.path.toLowerCase());
    };
```

**Edit2:**
I reverted the function call and it did it - sorting order is reverted ...

``` javascript
// fill the albums from Gallery.images
Gallery.fillAlbums = function () {
    var sortFunction = function (b, a) {
        return a.path.toLowerCase().localeCompare(b.path.toLowerCase());
    };
```

oparoz:
Fixes #90
A first step for #91


deMattin:
I like your idea to grey out (opaque) inactive sorting option and also button icons are ok now, I think.
Grouping the two buttons is also a good idea.

But why have you moved them to the left at the end of the breadcrumb?
On the left side there's the path (breadcrump) and the switch to the files app - that makes sense.
But sorting order is more some kind of an option, I would like to see right where the info and sharing buttons are.
And with walking through folders these buttons now walk over the screen with the right side of breadcrump, too.

I would definitly vote for removing the buttons to the right again.

oparoz:
The rationale is: the left side is for everything related to files, the right side for other types of actions

What do you mean by "And with walking through folders these buttons now walk over the screen with the right side of breadcrump, too." ?
Those buttons will always be next to the Files button, so they move with it. I don't see them "jumping".

deMattin:
They don't jump (no issue there). But they are not always at the same place.
Now I understand you intention, why you moved the buttons.

With the switch to the files app, this is ok and probably the best place, because you don't simply switch to files app, but you switch exactly there in the app folder, where the button now is placed.
So I think, it's intuitive with the files app switch button to place it there next to broadcrump.

But the sorting option is not a switch on patch - it's a switch on and in gallery showing a different view of the previews.
So in my opinion it's a gallery related option and not a file or folder related thing.
And also knowing and understanding your thoughts, it's a bit retarding the workflow, if the buttons are not always in the same place, too.

So all in all for me it would be more intuitive to bring it on the right again.
Maybe I am already used to the "right" place, too. But I don't think, this is the reason, why it seems to me a bit misplaced on the left.

oparoz:
Breadcrumb, link to Files, files sorting are all files operation to me.
Sharing, information are not.
At one point I thought about leaving them in the middle, as separate items, but it did not feel right.

We need more feedback, to see if it's just because you're used to it or if it feels wrong for everybody.

deMattin:
Agree, that we need more feedback - although it's harder to find something a better solution, if you don't have the possibility to directly compare. ;)

Mostly it bothers me, that the buttons walk with the breadcrump and have no fixed position in ownCloud workplace.
It is a normal workflow, that you work with an app without really analizing UI elements, if you use this app more often.
This is also the reason, why "powerusers" of an app mostly don't like such things like "ribbons", that change content on different situations and often mostly address the needs of beginners.
After a while you click on the 2nd button on the left side and not "search" a button, that looks like this or that ...
So it is stopping the flow, if you have to "visually search" for the place of a button.

It's hard for me to explan - even harder in english ...

oparoz:
I see what you mean, but I never had any issue finding the buttons while testing. You can immediately see that there are two groups of buttons at the top.
And sorting is not something people should use in every folder. I see it as more of a convenience tool.

deMattin:
hmm - for me it's the most used button in gallery ... ;)
Don't misunderstand me - it is useable now as it is and for sure I also find the buttons, but the "old way" seemed to be better and smoother in my opinion
But let's see, what others say.

I will have a look at it, how to make this part of css code more easily themeable.
As it is now, my 1st approach hacking css in browser had no success and the buttons go right most.
There seems to be also some alignment issues (I think, share and sort buttons alignment have to be defined separately in css).

Maybe it is a solution for all needs, if the position of the buttons will be made easy themeable?!

oparoz:
If you want them on the right, you need to hack the template as well. The order is reversed.
My main problem is that I don't have access to many owncloud users, so my testing panel is not great.

Lots of things will be possible if we get a user settings section one day.

PR #139, opened 2015-05-13
fix version compatibility so it's installable on ownCloud 8.1 / dev setups

jancborchardt:
Please review @oparoz @DeepDiver1975 

I found two things though:
- the minimum ownCloud version required is 8.0.2? Is that correct, so it doesn’t work with 7? I guess we just have to leave those folks with the old Pictures app.
- it only supports IE11? If we really want to ship this as the default Gallery app, we _need_ to support IE8 … at least IE9. cc @MorrisJobke

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -36,6 +36,6 @@
 	<ocsid>169116</ocsid>
 	<dependencies>
 		<php min-version="5.4" max-version="5.6"/>
-		<owncloud min-version="8.0.2" max-version="8.0.99"/>
+		<owncloud min-version="8.0.2" max-version="8.1"/>
 	</dependencies>
 </info>


jancborchardt:
I mean – maybe it works with IE10, 9 and 8 and just wasn’t tested? If not is it blocked hard or can we fix it? We kinda need to ;)

oparoz:
### Branches

Here is how it works
- `master`: that's where the development is done, targeting the latest stable version of ownCloud so that the most users can benefit from enhancement, unless a change only applies to the next release of oC
- `stableX`: that's where things end up after a period of testing on master and that's what is being used for the packaging of the app for the current version of oC
- `dev`: that's for things which only apply to the next release of oC

So, it's normal for master to target 8.0.2 to 8.0.99 since everything is designed for 8.0. stable8 gets the same file.
dev targets 8.1.0 to 8.1.99 and will be branched to stable8.1 once 8.1 is launched.

There is a stable7 branch which works with 7, but it doesn't contain any of the recent optimisations, so it won't work well with large collections. Since the architecture and the minimum PHP versions are different, nothing was backported.

### IE

Regarding IE, as I mentioned a few times, ownCloud uses HTML5's EventSource which is missing from Microsoft's family of browsers, so custom code has been developed for core, but it doesn't work at all as soon as there is more than one stream running at the same time (and maybe on other occasions). **To me that makes ownCloud incompatible with IE**, but with polyfills, it's possible to get Gallery working on IE11. I haven't tested on older versions.
So there is nothing blocking older browsers, the app is just encouraging users to migrate to Firefox or Chrome. If someone wants to improve compatibility, I'm happy to provide guidance on paths to explore.

jancborchardt:
@oparoz regarding the branches we should use the same branching system as core and most other apps:
- master is the dev branch, but also usable. The code always targets latest core master, in this case to-be 8.1
- stable8 is for 8.0.*
- stable7 for 7.0.*
- etc.

Much simpler, and if a fix needs to be backported we can just do it. :)

cc @DeepDiver1975

oparoz:
@jancborchardt - I'm a bit reluctant, because in reality, with other apps, only master gets some proper love and that's always for the next release. The current stable version only gets back ported bug fixes, just like in core and that goes against the principle of independence where apps evolve at their own rhythm.

Of course, now that it's in the owncloud domain you can do what you want with it :) and it might be best to follow that structure so that anybody can easily contribute, but I find cherry picking to be a lot more work than rebasing.

jancborchardt:
> The current stable version only gets back ported bug fixes, just like in core and that goes against the principle of independence where apps evolve at their own rhythm.

That’s how it’s supposed to be. People expect their ownCloud to work and not to break between, say, 8.0.3 and 8.0.4 because a shiny new feature was introduced. ;) That’s why _only_ bug fixes are backported.

We have a 3 month release cycle so the »proper love« will arrive for everyone soon enough. :)

Sure it might be a bit more work (or at least you need to change the way you do it a bit) but it is _very_ important so users get a stable and trustworthy product which works. And that, from experience, is _much_ more important than any feature.

oparoz:
I fully understand the philosophy for the development of core, but my understanding was that apps would be moved outside of core so that they could evolve at their own pace, independently. I think that's even in a video from last year's conf.

If not, let's take the example of an OS and it's apps.
Following the same principle would mean that you only get new apps or new features in apps when a new OS version is released.
That's rarely the case, the only exception being Linux in the early days, but mainly because a lot of the big apps where dependent on the DM's libs.
Nowadays, only the LTS version are stuck with old versions of apps, for stability reasons and that's perfectly understandable for server or enterprise products, but not really for consumers.

Back to ownCloud, if Gallery 8.0.20 introduces a change which unfortunately breaks the app for users with external storage turned on, they can simply install 8.0.19 and wait for the fix. It's also possible to have betas for releases which contain new features, etc.
I think it's much more beneficial to the end user than having to wait a few months before being able to use small improvements.

jancborchardt:
@DeepDiver1975?

schiessle:
> I fully understand the philosophy for the development of core, but my understanding was that apps would be moved outside of core so that they could evolve at their own pace, independently.

That's completely right and nobody stops you from releasing a new version every months, every 6 months or whatever fits best for your app. Still I think it would be nice to follow the general workflow we established for core and other apps:
- master: your dev branch with the latests code
- stable7, stable8,... for the code compatible with the corresponding ownCloud core. 
- if you have multiple versions within one ownCloud branch you could identify them by tags.

What I find confusing is that you have two branches with the identical code (master and stable8), this looks wired to me. Just a few minutes ago I stumbled across this problem because I checked out master and was surprised that it doesn't work with core master.

Just my 2 ct

oparoz:
Thanks for the feedback. My main problem is with PRs, which will only be made against master and will be harder (for me) to backport if master contains radical changes, but if 99% of apps follow the same model, then obviously it's much easier for everybody if I follow the model defined in core...

Branches are described here: https://github.com/owncloud/galleryplus/wiki/Branches

oparoz:
Another problem is around making it easy for contributors. If contributors are on the stable branch and want to help by providing patches compatible with their setup, they won't want to have to setup a copy of ownCloud master which may contain different methods, etc. It's a barrier to entry. I know it was for me. Having to fix the next version, hoping for a backport when I need to fix the version I'm using today.

jancborchardt:
So I guess this PR is obsolete and we’ll continue the discussion in https://github.com/owncloud/galleryplus/issues/209 :)

Issue #144, opened 2015-05-18
Add "Add to owncloud" button


oparoz:
As discussed in https://github.com/owncloud/core/issues/16386, it would be good to use the same header as the Files app when sharing folders.

One of the tasks is to add the "Add to owncloud" button.

PR #146, opened 2015-05-20
Adds the form which allows a user to add the shared files to their ownCloud

oparoz:
Solution for #144

### DIFF
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -68,7 +68,8 @@ class Application extends App {
 				$c->query('Request'),
 				$c->query('Environment'),
 				$c->query('OCP\IURLGenerator'),
-				$c->query('DownloadService')
+				$c->query('DownloadService'),
+				$c->query('OCP\IConfig')
 			);
 		}
 		);
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -16,6 +16,7 @@ namespace OCA\GalleryPlus\Controller;
 
 use OCP\IURLGenerator;
 use OCP\IRequest;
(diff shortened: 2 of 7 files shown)


oparoz:
@jancborchardt @libasys @nclm 

I've only tested this on OC8 so far.

oparoz:
Also @deMattin

jancborchardt:
cc @PVince81 @schiesbn for the server-to-server sharing review

oparoz:
So, any problem with the approach @PVince81 @schiesbn ?

Is it working for you @demattin @libasys @setnes @nclm

schiessle:
I tried the server-to-server share part... Works :+1:

oparoz:
Thank you!

jancborchardt:
@oparoz please merge https://github.com/owncloud/galleryplus/pull/139, otherwise I can’t test. ;)

oparoz:
I can't merge that PR. Branches have to be renamed if the way forward is to mirror the core setup.

If you're stuck on 8.1 though, you can use the ultimatedev branch which contains all the latest PRs.

jancborchardt:
@oparoz ok, it doesn’t have to necessarily be merging that branch. The point is that `max-version="8.0.99"` is not a proper attribute, and it should be `max-version="8.1"` ;)

oparoz:
Ah, no, this is to say that it can be installed on all 8.0.x version, but not 8.1.x, assuming development versions of 8.1 are never named something like 8.0.9.1

jancborchardt:
@oparoz delete merged branches please ;) press the »Delete branch« button here.

Issue #145, opened 2015-05-18
Turn the Gallery button into a switch


oparoz:
It would be more immediately understandable if there was a switch with both the Gallery and Files button, as shown here:
![0ebbfcfc-fcdb-11e4-877c-873b19def287](https://cloud.githubusercontent.com/assets/323220/7682644/dc28dbbc-fd79-11e4-9984-2fd249e208c9.png)

oparoz:
Change has been rejected. Will keep in Gallery+ for 8.1 and 8.0

jancborchardt:
@oparoz sorry again for swaying back and forth on this one. Oftentimes design has to be carefully evaluated as you know. :) Especially when it’s that present in our main Files app.

oparoz:
Yep, no hard feelings :)
Since 8.0 and 8.1 have the old design, it will work there and I'll be able to get some feedback :)

Issue #67, opened 2015-03-07
Not able to switch to gallery+ in (sub)folder with no pictures


deMattin:
It's a great feature to switch to gallery+ from file view in subfolders and start preview there.
I have noticed, that if you are in a folder where no pictures are in but subfolders with pictures, the switching button is disabled.

I think, it would be better, not to disable this switchbutton.
So you will be able to use every subfolder as a startpoint for gallery+.

oparoz:
Yes, I did notice that behaviour when having a root folder with no pictures per example, but thought it was OK as the menu item "Gallery+" would let the user go there anyway, but that doesn't work for your case.

I don't think there is a way to know what's in the sub-folders without looking, so there are 2 solutions :
1. Always keep the Gallery view button in the UI, but that means that in some cases, you will end up in an empty gallery, if there are no pictures in sub-folders. Maybe @jancborchardt and his team knows what's best here. Maybe having a consistent UI is best and users wouldn't be surprised by the behaviour I've just described. _Very light development work_
2. Query the server to see if there are pictures in the sub-folders. That's an extra request at every folder change. Does it matter? Are there too many already? _Medium development work_

deMattin:
I think,  it would really be the best, to **never** remove the switch and I see no disadvantages:
1. it is consistent
2. if you switch and there are no pictures in this or subfolders, you get a clear message
3. if pictures are in subfolders you get gallery+ working from this point in folder structure
4. no performance issues or performance disadvantages at all

oparoz:
I agree, but the contextual button is an old feature from the original Gallery project, so I'd love to get some feedback from people who've been involved with the project from the start.

deMattin:
I think, the old gallery didn't have this issue.
Isn't there the folder preview saved as a separate file?
But it's also possible, I mix it with the gallery enhancement of libasys (OC 7) which was on a good way but he doesn't continue to work on as it seams to me.

oparoz:
It's right here: https://github.com/owncloud/gallery/blob/master/js/public.js#L9 :)

I hope libasys hasn't given up and that' he'll contribute some of his frontend enhancements

oparoz:
I'm going to always enable it and see if people complain ;)

oparoz:
It's in

deMattin:
I like it very much!
The possibility to start gallery in any folder without pictures (but possibly with pictures in subfolders) is very helpfull.
And I see no bugs or disadvantages.

[...]

PR #148, opened 2015-05-20
Turn the Gallery button into a switch

oparoz:
Implementation of #145 

PR for ownCloud 8.0

@jancborchardt @libasys @nclm @deMattin, please test.

### DIFF
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -144,6 +144,25 @@
 			// Refresh the view
 			Gallery.view.viewAlbum(Gallery.currentAlbum);
 		},
+		
+		/**
+		 * Switches to the Files view
+		 *
+		 * @param event
+		 */
+		switchToFilesView: function (event) {
+			event.stopPropagation();
+		
+			var subUrl = '';
+			var params = {path: '/' + encodeURIComponent(Gallery.currentAlbum)};
+			if (Gallery.token) {
+				params.token = Gallery.token;
+				subUrl = 's/{token}?path={path}';
+			} else {
+				subUrl = 'apps/files?dir={path}';
+			}
+			OC.redirect(OC.generateUrl(subUrl, params));
+		},
 
 		/**
 		 * Populates the share dialog with the needed information
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -164,7 +164,6 @@
 			
 			var button = $('#filelist-button');
 			button.addClass('loading');
-			button.find('img').attr('src', OC.imagePath('core', 'loading-small.gif'));
 			OC.redirect(OC.generateUrl(subUrl, params));
 		},
 
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -154,7 +154,7 @@
 			event.stopPropagation();
 		
 			var subUrl = '';
-			var params = {path: '/' + encodeURIComponent(Gallery.currentAlbum)};
+			var params = {path: '/' + Gallery.currentAlbum};
 			if (Gallery.token) {
 				params.token = Gallery.token;
 				subUrl = 's/{token}?path={path}';
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -161,6 +161,10 @@
 			} else {
 				subUrl = 'apps/files?dir={path}';
 			}
+			
+			var button = $('#filelist-button');
+			button.addClass('loading');
+			button.find('img').attr('src', OC.imagePath('core', 'loading-small.gif'));
 			OC.redirect(OC.generateUrl(subUrl, params));
 		},
 
--- a/css/gallerybutton.css
+++ b/css/gallerybutton.css
@@ -1,14 +1,54 @@
-/* toggle for opening shared picture view as file list */
-#openAsFileListButton {
+/* Toggle for opening the files list in the Gallery app */
+#controls .button.view-switcher {
 	float: right;
 	margin-right: 4px;
 	margin-top: 5px;
 	font-weight: normal;
+	width: 40px;
+	padding: 8px 11px;
 }
 
-#openAsFileListButton img {
-	vertical-align: text-top;
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
-	filter: alpha(opacity=50);
-	opacity: .5;
+.button.left-switch-button img,
+.button.right-switch-button img {
+	vertical-align: -3px;
+	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=60)";
+	filter: alpha(opacity=60);
+	opacity: 0.6;
 }
+
+#controls .button.left-switch-button {
+	float: left;
+	display: inline;
+	border-radius: 3px 0 0 3px;
+	margin-left: 3px;
+	margin-right:0;
+	border-right: 0;
+}
+#controls .button.right-switch-button {
+	float: left;
+	display: inline;
+	border-radius: 0 3px 3px 0;
+	margin-left: -1px;
+}
+
+#controls .inactive-button {
+	background-color: rgb(230, 230, 230);
+}
+
+#controls .inactive-button img{
+	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=25)";
+	filter: alpha(opacity=25);
+	opacity: 0.25;
+}
+
+#controls .inactive-button:hover,
+#controls .inactive-button:focus,
+#controls .inactive-button a:focus{
+	background-color: rgba(255, 255, 255, .95);
+	color: #111;
+}
+
+#controls .disabled-button {
+	pointer-events: none;
+}
+
--- a/css/gallerybutton.css
+++ b/css/gallerybutton.css
@@ -37,26 +37,23 @@
 	opacity: 0.6;
 }
 
-#controls .inactive-button {
+#controls .disabled-button {
+	pointer-events: none;
 	background-color: rgb(230, 230, 230);
 }
 
-#controls .inactive-button img {
+#controls .disabled-button img {
 	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=25)";
 	filter: alpha(opacity=25);
 	opacity: 0.25;
 }
 
-#controls .inactive-button:hover,
-#controls .inactive-button:focus,
-#controls .inactive-button a:focus,
-#controls .inactive-button.loading {
+#controls .disabled-button:hover,
+#controls .disabled-button:focus,
+#controls .disabled-button a:focus,
+#controls .disabled-button.loading {
 	background-color: rgba(255, 255, 255, .95);
 	color: #111;
-}
-
-#controls .disabled-button {
-	pointer-events: none;
 }
 
 #share-button img,
--- a/css/gallerybutton.css
+++ b/css/gallerybutton.css
@@ -19,13 +19,15 @@
 	perspective: 1000px;
 }
 
-#controls .button.left-switch-button {
+#controls .button.left-switch-button,
+#controls .button.left-switch-button:hover {
 	float: right;
 	display: inline;
 	border-radius: 3px 0 0 3px;
 	margin-left: 3px;
 	margin-right: 0;
 	border-right: 0;
+	border-right: 1px solid rgba(216, 216, 216, 0.9);
 }
 
 #controls .button.right-switch-button {
@@ -46,7 +48,8 @@
 
 #controls .disabled-button {
 	pointer-events: none;
-	background-color: rgb(230, 230, 230);
+	background-color: rgba(230, 230, 230, .9);
+	border: 1px solid rgba(230, 230, 230, .9);
 }
 
 #controls .disabled-button img {
@@ -55,13 +58,27 @@
 	opacity: 0.25;
 }
 
-#controls .button.sorting.active-button,
 #controls .disabled-button:hover,
 #controls .disabled-button:focus,
 #controls .disabled-button a:focus,
 #controls .disabled-button.loading {
-	background-color: rgba(255, 255, 255, .95);
+	background-color: rgba(255, 255, 255, .9);
 	color: #111;
+	border: 1px solid rgba(240, 240, 240, .9);
+}
+
+#controls .button.sorting.active-button {
+	background-color: rgba(255, 255, 255, .9);
+	color: #111;
+	border: 1px solid rgba(240, 240, 240, .9);
+	border-right-color: rgba(216, 216, 216, 0.9);
+}
+
+#controls .button.sorting.active-button:hover {
+	background-color: rgba(255, 255, 255, .9);
+	color: #111;
+	border: 1px solid rgba(240, 240, 240, .9);
+	border-right-color: rgba(216, 216, 216, 0.9);
 }
 
 /* Original Flip CSS by David Walsh
@@ -89,7 +106,6 @@ http://davidwalsh.name/css-flip*/
 	-webkit-backface-visibility: hidden;
 	-moz-backface-visibility: hidden;
 	backface-visibility: hidden;
-
 
 	-webkit-transform-style: preserve-3d;
 	-moz-transform-style: preserve-3d;
--- a/css/gallerybutton.css
+++ b/css/gallerybutton.css
@@ -11,6 +11,9 @@
 
 #controls .button.sorting {
 	padding: 8px 7px;
+	-webkit-perspective: 1000px;
+	-moz-perspective: 1000px;
+	perspective: 1000px;
 }
 
 #controls .button.left-switch-button {
@@ -47,7 +50,7 @@
 	opacity: 0.25;
 }
 
-#controls .active-button,
+#controls .button.sorting.active-button,
 #controls .disabled-button:hover,
 #controls .disabled-button:focus,
 #controls .disabled-button a:focus,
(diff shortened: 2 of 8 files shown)


oparoz:
I'm thinking the transition could be refined by clearing the view and showing the spinner since the whole page has to be reloaded... No idea if that's easily doable or if that really improves the user experience.

jancborchardt:
Also cc @owncloud/designers, please review! :)

oparoz:
This time I'm happy with the way it works.

Please review on 8.0 @jancborchardt @jospoortvliet @libasys @nclm @deMattin @setnes

oparoz:
The ultimatedev branch for 8.1 contains this PR

jancborchardt:
I fixed the style for the buttons: It was confusing as to what state is the current one:
capture du 2015-06-29 16-49-39
Here for example I am in the files list, but the button definitely looks like I am in the Gallery view (button pressed and looks disabled). So I simply switched the styles around:
capture du 2015-06-29 17-03-06

jancborchardt:
Another thing: I would say we should put this switcher on the far right of the controls bar. Otherwise it takes focus away from the New & Upload buttons.

jancborchardt:
> We should put this switcher on the far right of the controls bar. Otherwise it takes focus away from the New & Upload buttons.

Done with the newest commit :)

oparoz:
I agree for the Files/Gallery switch, but it's going to be a problem with the sorting buttons since each button has 2 functions: activate the sort and change the order. Any idea on how to solve that?

Historically, that button has always been on the left, but I'm happy with the change if that's the new way forward :)

jancborchardt:
There’s only one problem now: The sort by date / name toggle doesn’t have a pressed state anymore. Can you add the .disabled-button logic to it because that’s the class we use now?

jancborchardt:
> activate the sort and change the order

What do you mean by »activate the sort«? Isn’t the default sort by date?

oparoz:
The default is by date, but if you click on the button again, you change the order. If that button is marked as disabled, then you can't do that any more and if it's greyed out, people won't think of using it.

oparoz:
A 3rd button to change the order is always an option, but it's starting to get crowded up there... No?

oparoz:
A dropdown is another option, but that means a minimum of 2 clicks for every change, unless we use a hover.

jancborchardt:
> The default is by date, but if you click on the button again, you change the order.

Aaah right! Yeah, a third button would be too much. I’m thinking if it would be better to have _one_ button for sorting in there which drops down a menu where you can choose:
- Newest first
- Oldest first
- Alphabetically

I don’t think that’s too bad regarding the 2 clicks because it’s something which is not changed so often, eh?

jancborchardt:
The other question is: Do we _really_ need reverse chronological order?

oparoz:
I think we do need the reverse chronological order. It's very useful on large albums, when you want to reach the oldest pictures right away.

For me there should be either 3 or 4  buttons, but since it's in a dropdown, we could use 4, saving another click.

And 2 clicks ins't that bad since the order can be saved in the album in case the default doesn't work for the owner of the library. Thinking about it, this could even be done from the dropdown, using a save tick box.

jancborchardt:
The sort order chosen should of course be remembered for that folder, and maybe all subfolders (the default being Newest first in every folder though).

oparoz:
Let's deal with the saving and default parts in separate PRs
https://github.com/owncloud/galleryplus/issues/197
https://github.com/owncloud/galleryplus/issues/198

oparoz:
So, 3 items left in this PR
- [ ] Restore View switcher in the Files view. Float:right kills it.
- [ ] Dropdown UI with 4 buttons for sorting
- [ ] Connect the logic to the new UI 

I think we should also give some thoughts to the buttons order. It's currently
Share - Info - Sort - View
and it feels a bit confusing (one reason I kept 2 groups)

oparoz:
@jancborchardt - I've lost this branch locally. I can restore it from backup, but it won't have your latest changes as I had forgotten to pull them.

I think the best thing to do would be to re-create a PR in owncloud/gallery once g+ is the new master.

jancborchardt:
You lost it, how that? :D See https://github.com/owncloud/galleryplus/pull/214 :D

oparoz:
Yeah, I wanted to cherry-pick the changes, but when moving things around I did not notice that I only had my commits in my local branch :facepalm:

oparoz:
Ok @jancborchardt , we're back in business and we need you to fix the first 2 of the list https://github.com/owncloud/galleryplus/pull/148#issuecomment-116769686

oparoz:
@jancborchardt - Do you have time to fix this by the end of the week? If not, I think I'll split it in 2
- turn the button into a switch
- redesign switches and change button placement

jancborchardt:
Let me have a look at the first one. :)

oparoz:
@jancborchardt - Change of plan.

This version includes everything. I'm simply using the hover style to indicate which sort button has been selected. Another style can be used of course, if you have a better idea.

Bonus: The active button now flips between states.

The biggest thing left is for you to fix your changes on the Files side. The switch disappears on IE9 on master and there may be other issues on other browsers and with older versions of ownCloud. 
In short, the switch doesn't like to be on the right in the Files app. There must be some CSS conflict.

Then we can talk button order and merge this.

Step 2 will be to come up with a better sorting interface, with a dropdown, as discussed above.

I would also like to get some other people to take a look at this: @libasys @nclm @deMattin @setnes

deMattin:
Recently I had problems to setup git in my owncloud (conflict with check for existing gallery app or blank screen - maybe related to actual git changes) so I had to switch to official app release in app store - I will have a try later to get in touch with actual development branch.

I just want to give my thoughts about a pulldown for sorting:
In my point of view, this is the most often used "switch" for a folder related switch setting and I use it often in normal workflow. In files app I change sorting very rare!
So I _normally_ would prefer to stay on some switch buttons and don't hide this in a pulldown, which will use one more click to be used or for change this setting,

But a Button, indicating the actual sorting order and giving a pulldown on click with all possible sorting orders, will also fit well.
And switching to a reverse sorting is very useful in a gallery or photo app!

libasys:
looks nice but a little bit to much flip on the sorting buttons! Button switch in the files app, is that the default behaviour,  when there are no images in that folder the switch is active? I think if there are no images then the button should become inactive!

oparoz:
@demattin - I fixed the blank screen problem, so both can be installed at the same time, but that won't be good for testing :). Since you have your own fork, you may need to reset the upstream link. There is plenty of doc, but you can always ask on IRC and people will help you.
And thanks for the feedback regarding sorting.

@libasys - What do you mean by "too much flip"? Maybe I need to adjust the speed.
And yes, the switch is always active, because there could be pictures in sub-folders (https://github.com/owncloud/gallery/issues/67)

deMattin:
1. The flipping is a nice idea to show, where a click does toggle to. I think, it's intuitive (I understood it immediately). It's a good approach until pulldown is worked out.
2. Sorting and other buttons on the right is a thing, I like more (oparoz, you remember our discussion a long time ago... ;)).
3. The switch-to-gallery-button in files app is centered now. Doesn't it has to be aligned on the right side?
4. I don't like the presence of the gallery switch button (greyed as inactive) on the right in gallery. Before it has been better, when the switchbutton changes between switch-to-files and switch-to-gallery.
5. Switch button active in files view even if there's no media is ok, I think. It's better to get a message, that there's no media file, than to see this button coming and going while you navigate. And second it will slow down files app, if every folder (and subfolders!) is checked in files app. This shouldn't be done!
6. Switching back to files app in a folder with no media does not work.

@oparoz: I didn't have installed both or 2 different galleries. But neither galleryplus nor gallery folder worked for my git branch. The one path gave a blank screen and the other an error message, that another gallery has been installed. Seems, that something got messed with this renaming of the app (in my install). Now it works!

@oparoz: Which git is the one for development? Your master in "interfasys/galleryplus" or the master in owncloud/gallery?

oparoz:
- 3) That's a bug @jancborchardt needs to fix
- 4) Once the Files thumbnail view has been implemented, then a switch looking just like the one in the PR should be used to quickly change view, but there would be an extra, optional Gallery button to allow users to completely change app.
  Since that view doesn't exist yet, having the switch makes it easier to understand that there is an alternative way to look at the current folder. I see this as a temporary solution
- 6) I'll take a look. I did notice that the Files app got reset sometimes and I did fix that once, so I need to figure out if something got lost or if there are different cases...

@demattin - I think you got stuck because some of the branches got deleted. If you had created your own fork, you can simply change the origin and upstream urls and should be OK after a rebase.

`owncloud/gallery` is for 8.2
`interfasys/galleryplus` is for 8.1/8.0

deMattin:
If `owcloud/gallery` depends on oc8.2, point 6) may be caused by my oc8.1 using this branch?
`owncloud/gallery` Master is working fine with oc8.1.
I think as long as `owncloud min-version` is set to 8.1, it shouldn't be a problem to use this branch with oc8.1.

Btw. (found it again by searching min-version):
Please delete double line spacing in `<description>` tag in info.xml. The spacing line after every line doesn't really look good and no other app description looks like this.

oparoz:
I can't delete the double space in the description. Without it, everything collapses and you can't read a thing. It's a problem with the way the descriptions are parsed which will hopefully be fixed one day when the app manager supports markdown.

jancborchardt:
(@oparoz regarding the double-linebreak, can you please open an issue in core? :)

oparoz:
@jancborchardt - It's been open for close to 2 years: https://github.com/owncloud/core/issues/5322

js/gallery.js
```diff
@@ -144,6 +144,28 @@
...
+		/**
+		 * Switches to the Files view
+		 *
+		 * @param event
+		 */
+		switchToFilesView: function (event) {
+			event.stopPropagation();
+		
+			var subUrl = '';
+			var params = {path: '/' + encodeURIComponent(Gallery.currentAlbum)};
```
oparoz: That's where bug 6) is located. It's a bad merge from js/app.js

oparoz:
So @jancborchardt , when will you be able to fix the buttons' CSS? :)

MorrisJobke:
> So @jancborchardt , when will you be able to fix the buttons' CSS? :)

@jancborchardt Isn't conf the best time to work together on this with @oparoz ;) 

Also a rebase would help here.

jancborchardt:
@oparoz so anything here missing from your side? Otherwise we should just rebase, right?

jancborchardt:
Or what do I need to do?

oparoz:
1. Rebase
2. If the rebase requires lots of merges, then I'll do it
3. If not, then simply refer to https://github.com/owncloud/gallery/pull/148#issuecomment-126154147 It's about the switch button on the Files app

I'll take a look after your design workshop.

oparoz:
OK, it's been rebased.

The recent changes in core master make it so that the switch seems to work fine now on the Files side.

There is one problem remaining. The sharing button disappears on Firefox. Data looks really strange in the debugger, so I'm not sure what to make of it.
I'll let @jancborchardt figure it out :)

jancborchardt:
After trying it out more and also talking to @karlitschek I decided that a toggle is not desirable here as it adds way too much interface elements. It’s a giant toggle with lots of attention. cc @owncloud/designers 

I will slightly change the icon and adjust the style so it fits more in the bar. We can use the right floating of the sorting toggle from this PR, but maybe it makes sense to rather start newly.

oparoz:
After having used both metaphors, I think a switch makes it easier to understand what the button is about and eventually you're going to add one when introducing the grid view, but in the mean time, I'm not dead set on that particular solution. I think the better approach is to find a way to group some of the other buttons and to use the right side-bar.

Every file organising system in the world has buttons in the controls bar. https://support.apple.com/en-us/HT201732
example

karlitschek:
@oparoz Yes. A switcher might be the best solution once we have the grid view. But until then (aka 8.2) I would suggest to do the simpler version as @jancborchardt proposed.

oparoz:
@jancborchardt - Please provide a quick mockup before we start working on the code. The following items will all end up on the right side if that's the way forward:
- share button
- info button
- sorting switch
- Files view button

They need to be in some sort of logical order and probably split in groups.

jancborchardt:
@oparoz sure:
- share button: Should be in the breadcrumbs, on the right next to the current folder name
- Files view button: to the far right in the controls bar, like now.
- sorting toggle: if there would be a settings area, that should go there. Since there is none for now – let’s simplify it by making it a simple toggle: Either sorting by date (newest first) or alphabetically. The other two options make less sense and just add huge complexity to the interaction. The single button should be on the left of the files view button.
- info button: not really sure about this one. If there’s text set for the current folder, I would not only show it inside of this tiny popover, but rather properly render it above the pictures. What do you think @oparoz?

deMattin:
**PLEASE don't kill "reverse date+name sorting" (fast button) switch!**

I really don't understand why there's a problem with a handfull of buttons in an app and why there's a need to crop features only to have less buttons.
There's really no overload of features or visible buttons there!

And (probably just a mistype):
- share should go to the right
- switch to files app should go behind breadcrumb (like it is in files app)

And btw.:
It's a viewing app!
So kill the share feature buttton but not viewing relevant features, if you think, you have to kill buttons!

oparoz:
@jancborchardt 
- OK, so share is really on the left, on its own.
- The sorting mechanism is not more complex than what Files is doing and having it as a switch makes it act more like one button (control), although the new borderless design make it less obvious
  Given the different use cases for the gallery albums, I don't think we can pick one sorting order or another
- Regarding the info button, content can't always be spread horizontally and as soon as you have a list per example, there is going to be a large empty area on the right. Maybe we should use the right side-bar, but the problem is that there would still be a button to hide/show it

@deMattin When looking at the UI when using this PR, you notice a lot of buttons on the right and they all look quite the same, just with different icons. That's one reason I have currently 2 groups of icons in the app. info/action on the right and sorting on the left
So, something needs to be done and we're exploring options right now, but there will always be Gallery+ as a fallback.

jancborchardt:
@oparoz how about for the sorting we use the dot-dot-dot icon and expand it to show an action menu much like in files? Then we can display all the sort options there. If you think that we can’t display the info nicely directly in the body of the gallery (which I think we could do somehow and it would look more awesome) then we can also move that into this menu.

oparoz:
Ah, the 3 dots, I need to file a couple of issues about that... I'm not sure it would work with sorting. It would work well to put all sort of actions though, like rename, delete, once we're ready to add them.
In a files management solution, you don't want sorting to be hidden away and I would assume it to be the same when viewing photo albums. On top of that, we would need to show all 4 options in that action menu.

We could go back to one of the solutions described in this thread. Having one sorting button which would reveal all 4 options. I prefer my switch with flipping icons, but I could be convinced to change my mind if it looks awesome and is accepted by all.

If you want to play with the info text, take the one in the demo as an example
https://oc8demo.interfacloud.com/index.php/apps/galleryplus/s/4jjwPv5E21l8Lnk

I don't think we can make it fit nicely above the photowall.

jancborchardt:
> In a files management solution, you don't want sorting to be hidden away

In Files the sorting is nicely integrated, part of the header. In Gallery this is not possible.

Also, sorting is not something which is used all the time (unlike let’s say the view switcher or simply navigating the photos), so it’s fine to put it in a submenu. I would not call that hiding away, but rather it helps to not distract from more core aspects of the app.

jancborchardt:
> We could go back to one of the solutions described in this thread. Having one sorting button which would reveal all 4 options.

This is exactly what I mean. Whether we use the dot-dot-dot icon or some other icon is a detail. :)

oparoz:
> This is exactly what I mean. Whether we use the dot-dot-dot icon or some other icon is a detail. :)

No, I mean the sorting buttons need to be in their own dropdown, separate from actions like rename, delete, etc. The alternative being to use the 3 dots and to have sections in that dropdown

```
------------------------
Sorting
| au | ad | du | dd |
------------------------
Actions
| ed | del |
------------------------
```

Anyway, we need mockups to be able to start to understand what it will all look like

oparoz:
Since this PR is about turning the Gallery button into a switch, I think we should look at hiding the sorting buttons in another PR. Also, it has now been decided that there will be no switch-type buttons for files/gallery in Gallery, so that needs to be fixed.

Left to do here:
- [ ] Turn the switch-type buttons for Files/Gallery into single buttons
- [ ] Move share button to the left. @oparoz
- [ ] Make the button on the Files side transparent. Separate PR @jancborchardt 

Do we agree @jancborchardt ?

jancborchardt:
> Also, it has now been decided that there will be no switcher for files/gallery in Gallery

@oparoz huh, when has that been decided? :)

oparoz:
@jancborchardt - Wrong choice of words :D . What I mean is that it will go back to be a single button instead of two in a switch disposition.

oparoz:
I did some experiments and moving buttons around just doesn't work, because of the dropdowns. Clicking on share and getting a dropdown on the right doesn't work. Same for the info button. Moving the dropdowns to where the buttons are doesn't work. You have this large thing coming down in the middle which looks out of place. It will be best to move them when the right sidebar is implemented.

So for now I'll only move the sorting buttons, either to the left or to the centre of the controls.

oparoz:
@jancborchardt - OK, so it's rebased, you can break the switch here if you want or in a separate PR.

karlitschek:
@jancborchardt @oparoz any chance to have to gallery button improved in the files view?
We need this today.

oparoz:
Will rebase and merge so that @jancborchardt can then focus on the Files side.

jancborchardt:
From my side only https://github.com/owncloud/gallery/pull/311 is necessary, and I still think we should merge that instead of relying on this very large pull request first. We can then rebase this to go forward.

@oparoz we have core feature freeze today so please make sure we merge https://github.com/owncloud/gallery/pull/311 today – or if you choose to merge this, make sure it looks like intended by https://github.com/owncloud/gallery/pull/311

karlitschek:
yes. would be great to get this in. The rest can be done for the next release.

oparoz:
#311 cannot be merged, it doesn't look right on the Gallery side



3 button styles in one app

oparoz:
This PR isn't large, it's just swapping things around and has a more generic way of building buttons.
I'll prioritise adding the #311 changes to the Files side only, but the Gallery side also needs to be fixed as there will still be 2 different button designs after that
- [x] Turn the switch-type buttons for Files/Gallery into single buttons
- [x] Make the button on the Files side transparent

jancborchardt:
@oparoz that looks totally fine with the view switcher.
- sharing should be in the breadcrumbs, as discussed (without the button style as well)
- the info icon won’t be there by default for most
- the sorting buttons, again, should be behind a dot-dot-dot icon-more icon.

oparoz:
The sharing button can't go on the left, see https://github.com/owncloud/gallery/pull/148#issuecomment-139615549

The sorting buttons can be hidden later. We just can't have a UI in transition and which looks broken to users.

oparoz:
## Files view

files_view

jancborchardt:
Cool! :) See, the files view looks much nicer now with the way less obtrusive switcher.

oparoz:
I'm still not convinced, but it's more difficult to make the switch work with a flat design which has no borders or colours and very little foreground/background contrasts. I would probably have to play with the size of the button or something, but I would love to see the full UI elements sheet at some point.
I'll test both button types on Gallery+ to see which works best, but for older versions I'll keep the switch since buttons still have borders.

For sorting, it's more complex since there are 3 states, so one of them has the same border as the hover, because white on white doesn't work. We'll see how that evolves.

## Gallery view

gallery_view

jospoortvliet:
I know it's painful... But the end result does look good. I guess - no pain = no gain, right?

karlitschek:
yes. this is a good solution. Thanks guys

PR #153, opened 2015-05-21
replace logo-wide on share page with better icon + text, as in core

jancborchardt:
According to https://github.com/owncloud/core/pull/16517 – please review @owncloud/designers @oparoz @icewind1991 @DeepDiver1975

### DIFF
--- a/templates/public.php
+++ b/templates/public.php
@@ -44,11 +44,23 @@ style(
 </div>
 <header>
 	<div id="header">
-		<a href="<?php print_unescaped(link_to('', 'index.php')); ?>" title="" id="owncloud">
-			<div class="logo-wide svg">
-				<h1 class="hidden-visually"><?php p($theme->getName()); ?></h1>
+		<a href="<?php print_unescaped(link_to('', 'index.php')); ?>"
+		   title="" id="owncloud">
+			<div class="logo-icon svg">
 			</div>
 		</a>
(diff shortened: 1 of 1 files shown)


MorrisJobke:
Tested and works :+1:

oparoz:
Doesn't work. completely breaks the header on 8.0. The updated core CSS would need to be added to the app's CSS or the PR should be moved to sit on top of dev if the original change in core is not intended to be backported to stable8.

MorrisJobke:
> sit on top of dev if the original change in core is not intended to be backported to stable8.

The change is only intended for 8.1. @oparoz Can you cherry pick this into dev? @jancborchardt is on vacation this week. Thanks

oparoz:
OK, will do

oparoz:
It's merged

oparoz:
@jancborchardt - If you've submitted this patch to a few apps, you may want to make a modification so that they don't lose their compliance status. Your patch introduces a private API in the template.

Here is how I fixed it:
https://github.com/owncloud/galleryplus/commit/84141592e65f868955e5472dabc773571c5a34e1

jancborchardt:
@oparoz right, thanks for the note! @schiesbn @DeepDiver1975 can you comment on that? ^ Is it correct and should we do it like that?

Issue #157, opened 2015-05-23
Too many preview images are generated in the thumbnail directories.


setnes:
Preview images are generated for the size of the browser window.  For example, if someone views images with a browser sized at 1355x950, all of the preview images are are created at that size.  Later, if someone views images with a browser sized at 1349x943, all of the preview images are created again at this new size.

Also, preview images are generated for image sizes larger than the actual size of the image even if preview_max_scale_factor is set to 1.  This results in many duplicate files of the same resolution when the image is smaller than the window sizes.  Instead, we should serve the *-max.png in these situations instead of creating additional previews.

All of these extra preview images cause unnecessary overhead and storage on the server side.  We should create preview images that are **close to** the browser size instead of **exactly** the browser size.

getPreviewUrl in galleryutility.js can be adjusted to change the requested image size to normalize the requests to image sizes that have a greater probability of existing.  This is discussed under #54.

setnes:
The following code finds the smallest width and length dimensions that are multiples of 100 and are still greater than the window size of the browser.  This was tweaked to handle the case where the window size is, in fact, a multiple of 100.  My original statements would give 100 pixels more than it needed in that situation.

This goes in getPreviewUrl in galleryutility.js.

```
/* Find the next larger image width and length. */
if ( width % 100 != 0 ) {
      width = ( width + 100 ) - ( width % 100 );
}

if ( height % 100 != 0 ) {
      height = ( height + 100 ) - ( height % 100 );
}
```

There's still room for improvement here.  An image size of 1000x500, could get requested at 1000x1000, 1000x900, 1000x600, etc.  All of those would still result in different preview images that have identical actual resolutions.  We need some aspect ratio, longest edge comparison logic.

setnes:
This isn't quite right yet.  The image size requesting works fine, but the slideshow is not set to have the browser scale anything.  How can that be controlled?

oparoz:
I completely agree with the OP. It was on my todo list to fix core so that it wouldn't create previews larger than the max size or the original size.

So here is the todo:
- [x] Never create a cached preview larger than the max preview https://github.com/owncloud/core/issues/16545
- [x] Don't store the exact same preview every time we check for its existence https://github.com/owncloud/core/issues/16544
- [ ] Create preview categories https://github.com/owncloud/core/issues/8196
- [x] Fix the slideshow to make it possible to use previews which are larger than the viewport

[...]

Issue #142, opened 2015-05-16
Resizing browser window or rotating mobile device should result in corrected image size being displayed.


setnes:
In the gallery slideshow, if a user changes the browser window size, the image should be corrected to fit the new window size.  This is very evident when using a mobile device to view photos.  It's normal to assume a user will try rotating a phone to match portrait or landscape images and expect the image to fill the screen correctly.

This might mean we should request the scaled image based on the longest edge of the screen for both X and Y and then always resize to fit the browser.  Otherwise the image might need to be re-downloaded (bad?).  I'm not sure.

Again, PhotoSwipe does this correctly, but there might be other ways to accomplish this.

http://photoswipe.com/

oparoz:
I have an idea on how to quickly solve this...

oparoz:
After having toyed with it a bit it seems it's more complicated than I anticipated, so I will see if someone else wants to give it a go.

Photoswipe could be a better alternative to Bigshot since it supports zooming, swiping and resizing, but it's too early to tell and we would lose the proper zooming that Bigshot offers (not implemented yet).

oparoz:
I've had a look at Photoswipe's API and one reason it works well is because it needs to be given the dimensions of the images at initialisation time.
http://photoswipe.com/documentation/faq.html#image-size

Unfortunately ownCloud doesn't know anything about the images it's sending back, so there is no immediate solution, but starting from 8.1, oC will have a max size preview from which all subsequent previews are generated and we might be able to hack something together to calculate the image size based on that information and the screen size.
1. Get thumbnail
2. Looks at max preview
3. Send back thumbnail along with size meta data
4. Adjust dimensions to make them fit on screen
5. Ask for preview of calculated dimensions

Regarding the change of browser window size, past a certain point, there is no way around loading a new image. The core preview system should be upgrade one day to offer S, M and L previews and that would make things easier.

oparoz:
The above approach can't be taken for 2 reasons:
1. There may not be a preview yet, like on the Files app, with freshly uploaded images
2. We don't always have thumbnails before launching the slideshow

So we'll have to wait for a version of oC which supports meta data and automatic thumbnail generation on upload.

setnes:
I really appreciate the time spent looking into this issue and #141.  It sounds like both would be helped by some image meta data storage in the future.  Thanks again.

oparoz:
We need some people to test this on their mobile
@demattin @libasys @jancborchardt @georgehrke @jospoortvliet 

[...]

PR #159, opened 2015-05-25
Tweaks for fewer preview sizes and device rotation.

setnes:
Scale images in Bigshot that are larger than window.
Request preview sizes in multiples of 100.
Request preview sizes that still work when devices are rotated.
#157
#142

### DIFF
--- a/js/galleryfileaction.js
+++ b/js/galleryfileaction.js
@@ -62,6 +62,14 @@ var galleryFileAction = {
 		var width = Math.floor($(window).width() * window.devicePixelRatio);
 		var height = Math.floor($(window).height() * window.devicePixelRatio);
 
+		/* Find value of longest edge. */
+		var longEdge = Math.max( width, height );
+
+		/* Find the next larger image size. */
+		if ( longEdge % 100 !== 0 ){
+			longEdge = ( longEdge + 100 ) - ( longEdge % 100 );
+		}
+
 		for (var i = 0; i < files.length; i++) {
 			var file = files[i];
 			// We only add images to the slideshow if we think we'll be able
@@ -69,8 +77,8 @@ var galleryFileAction = {
 			if (galleryFileAction.mediaTypes[file.mimetype]) {
 				/* jshint camelcase: false */
 				var params = {
-					width: width,
-					height: height,
+					width: longEdge,
+					height: longEdge,
 					c: file.etag,
 					requesttoken: oc_requesttoken
 				};
--- a/js/galleryfileaction.js
+++ b/js/galleryfileaction.js
@@ -59,8 +59,8 @@ var galleryFileAction = {
 		var start = 0;
 		var images = [];
 		var dir = context.dir + '/';
-		var width = Math.floor($(window).width() * window.devicePixelRatio);
-		var height = Math.floor($(window).height() * window.devicePixelRatio);
+		var width = Math.floor(screen.width * window.devicePixelRatio);
+		var height = Math.floor(screen.height * window.devicePixelRatio);
 
 		/* Find value of longest edge. */
 		var longEdge = Math.max( width, height );
--- a/js/galleryutility.js
+++ b/js/galleryutility.js
@@ -100,11 +100,20 @@
 		getPreviewUrl: function (fileId, etag) {
 			var width = $(window).width() * window.devicePixelRatio;
 			var height = $(window).height() * window.devicePixelRatio;
+
+			/* Find value of longest edge. */
+			var longEdge = Math.max( width, height );
+
+			/* Find the next larger image size. */
+			if ( longEdge % 100 !== 0 ){
+				longEdge = ( longEdge + 100 ) - ( longEdge % 100 );
+			}
+
 			/* jshint camelcase: false */
 			var params = {
 				c: etag,
-				width: width,
-				height: height,
+				width: longEdge,
+				height: longEdge,
 				requesttoken: oc_requesttoken
 			};
 			return this.buildGalleryUrl('preview', '/' + fileId, params);
--- a/js/galleryutility.js
+++ b/js/galleryutility.js
@@ -98,8 +98,8 @@
 		 * @return {string}
 		 */
 		getPreviewUrl: function (fileId, etag) {
-			var width = $(window).width() * window.devicePixelRatio;
-			var height = $(window).height() * window.devicePixelRatio;
+			var width = Math.floor(screen.width * window.devicePixelRatio);
+			var height = Math.floor(screen.height * window.devicePixelRatio);
 
 			/* Find value of longest edge. */
 			var longEdge = Math.max( width, height );
--- a/js/slideshowzoomablepreview.js
+++ b/js/slideshowzoomablepreview.js
@@ -45,7 +45,9 @@
 			var maxZoom = this.maxZoom;
 			var imgWidth = image.naturalWidth / window.devicePixelRatio;
 			var imgHeight = image.naturalHeight / window.devicePixelRatio;
-			if (imgWidth < this.smallImageDimension && imgHeight < this.smallImageDimension) {
+			if ( imgWidth < this.smallImageDimension &&
+			     imgHeight < this.smallImageDimension &&
+			     this.currentImage.mimeType !== 'image/svg+xml' ) {
 				maxZoom += 3;
 				this.currentImage.isSmallImage = true;
 			}
@@ -57,9 +59,9 @@
 				width: imgWidth,
 				height: imgHeight
 			}), image);
-			if (this.fullScreen === null && this.currentImage.mimeType !== 'image/svg+xml') {
-				this._resetZoom();
-			}
+
+			// Reset our zoom based on image and window dimensions.
+			this._resetZoom();
 
 			// prevent zoom-on-doubleClick
 			this.zoomable.addEventListener('dblclick', function (ie) {
@@ -118,7 +120,14 @@
 			}
 			if (this.currentImage.isSmallImage) {
 				this.zoomable.flyTo(0, 0, this.smallImageScale, true);
+			} else if ( $(window).width() < this.zoomable.width ||
+				    $(window).height() < this.zoomable.height ) {
+				// The image is larger than the window.
+				// Set minimum zoom and call flyZoomToFit.
+				this.zoomable.setMinZoom(this.zoomable.getZoomToFitValue());
+				this.zoomable.flyZoomToFit();
 			} else {
+				this.zoomable.setMinZoom(0);
 				this.zoomable.flyTo(0, 0, 0, true);
 			}
 		},
@@ -180,8 +189,18 @@
 			}
 			if (this.currentImage.isSmallImage) {
 				this.zoomable.setZoom(this.smallImageScale, true);
-			} else {
-				this.zoomable.setZoom(0, true);
+			} else if ( $(window).width() < this.zoomable.width || 
+			            $(window).height() < this.zoomable.height ||
+				    this.fullScreen !== null ||
+				    this.currentImage.mimeType === 'image/svg+xml' ) {
+				// The image is larger than the window, or we are fullScreen,
+				// or this is an SVG. Set minimum zoom and call zoomToFit.
+				this.zoomable.setMinZoom(this.zoomable.getZoomToFitValue());
+				this.zoomable.zoomToFit();
+			} else { 
+				// Zoom to the image size.
+				this.zoomable.setMinZoom(0);
+				this.zoomable.setZoom(0, true); 
 			}
 		},
 


oparoz:
OK, I've now tested this branch and I've found 2 problems.
1. It's possible to zoom out too far. It was decided back when people were discussing the original Bigshot PR that it should not be possible to get a picture smaller than the original size. In this PR, since the picture is likely to be bigger, the limit can be set at the screen's dimension
2. When rotating panoramic pictures, they're not resized properly. They will be cropped and the user is able to pan.

setnes:
1)  Yes, I had to set minZoom to a value less than 0 to allow zoomToFit to downsize the image to the window size.  I might be able to compute a different value to prevent being able to zoom out farther than needed, but it would need to be computed for each image.  Right now it is a static -2... which can be drastic depending on the factors leading to the preview size.

2)  Actually, this is true for pictures other than panoramas as well.  I haven't figured out where we can call zoomToFit when the device is rotated, but before the next image is displayed.  What happens now is that the image is scaled properly when it is first displayed.  If you rotate the screen, it is displayed 1:1 until you navigate to another image and back.  The 1:1 display is larger than the screen (assuming the original image is also), so it only looks fugly if the aspect ratio is significantly different from the screen.

I'll try looking at 1, but I could use some thoughts on how to solve 2.

setnes:
Fixing number 1 was easy.  Bigshot has a method to call to compute the zoom factor.  Basically we can leave the minZoom defaulted to 0, then, right before we call zoomToFit, we can reset the minZoom to match the necessary zoomToFit value.  This way we can never zoom out (make the image smaller) than the window size... unless, of course, the image is smaller than the window already. 

I'll update the pull after I test a few other things.  The "getZoomToFitValue" method prevents me from having to figure out any math.  :)

```
// If the image is at all bigger than the window, call zoomToFit.
if ( $(window).width() < imgWidth || $(window).height() < imgHeight ) {
    this.zoomable.setMinZoom(this.zoomable.getZoomToFitValue());
    this.zoomable.zoomToFit();
}
```

setnes:
I'm going to move my logic to the _resetZoom function.  I see this is already being called in startBigshot.  Let me know if you don't want it there.

oparoz:
Regarding your second point, this works to call the proper function

```
var self = this;
window.addEventListener('orientationchange', function () {
    self.zoomToFit();
}, false);
```

but I need to see if this works with your solution since in master, zoomToFit will stretch smaller pictures

And for the rest, are you at ease with rebasing code? I had to touch some areas which affect your PR and so you would need to remerge it on top of master. It's really easy to do with an IDE, less so from the command line, except that I think that Github provides a visual tool for that... Anyway, let me know.

js/galleryfileaction.js
```diff
@@ -62,6 +62,14 @@ var galleryFileAction = {
 		var width = Math.floor($(window).width() * window.devicePixelRatio);
 		var height = Math.floor($(window).height() * window.devicePixelRatio);
 
+		/* Find value of longest edge. */
+		var longEdge = Math.max( width, height );
+
+		/* Find the next larger image size. */
+		if ( longEdge % 100 != 0 ){
```
oparoz: !== instead of !=

js/galleryutility.js
```diff
@@ -100,12 +100,21 @@
 		getPreviewUrl: function (image, etag) {
 			var width = $(window).width() * window.devicePixelRatio;
 			var height = $(window).height() * window.devicePixelRatio;
+
+			/* Find value of longest edge. */
+			var longEdge = Math.max( width, height );
+
+			/* Find the next larger image size. */
+			if ( longEdge % 100 != 0 ){
```
oparoz: !== instead of !=

oparoz:
OK, so if I call _resetZoom, I can see that it almost works. I need to call it twice... I think Bigshot is not using the proper dimensions when calculating the zoom factor.
It might be possible to do it manually.

oparoz:
I got it to work :), so there is hope for this PR :)

setnes:
I'm not able to work on this until I get home tonight, but I am curious
about the need to call _resetZoom twice.  It was working great for me last
night the way I had it.

I looked into the this.zoomable.registerListener logic the other day, but I
couldn't see where bigshot was raising any events for window size changes.
Is orientationchange a real event, or is it just something we wished was
there?  :)  Something is clearly happening though because any window resize
results in a 1:1 view.  There must be something to latch onto there.

I'll try to rebase the code in my pull tonight.  I'm new to Git, but
understand I need to merge some changes you made to the same files...
lines.. etc.  I'll be able to figure that out.

oparoz:
The code I posted earlier is all that is required to get it to work. My problem was that there are 2 events being fired when the device is rotated and Bigshot comes with default actions which obviously don't work properly when we trick it by playing with the image size.
So, the solution is to capture both events in the same call.

I'll try to add a commit to your PR, so that you get the full solution before the rebase.

oparoz:
I keep forgetting that I can't add commits to a fork...
I've added my commit to master, so you'll get it once you rebase :)

setnes:
I copied and pasted your "orientationchange resize" logic into my working ownCloud and I love it.  That was the missing piece of the puzzle.  This is SO much better on a mobile device now with all of these changes.  

I have other plans tonight, and I didn't get very far on rebasing on the command line.  I'm building git-cola right now.  I'll give it a shot later.  Part of my problem is this is the first time I've used git.  The other problem is I haven't taken the time to setup a proper test environment.  This is also why I haven't tested your core changes for previews.  I'm not even sure how to check those out.  Do I need to fork core even if I don't intend to make any changes there?

oparoz:
I agree :)

Let me know how you get on with using git-cola. The most important thing for a rebase imho is the diff tool. 

Regarding core, the most time consuming thing is probably setting up the database, but after that, you're up and running really quickly using git and it's easy to [switch to a PR](https://help.github.com/articles/checking-out-pull-requests-locally/).
The one thing I find annoying is that if you run tests, then your accounts are emptied, unless you create a skeleton which will be copied over at the end of the run.

oparoz:
There is also one last thing to fix. When the picture is larger than the screen, pressing down will resize it to it's original size, which is larger than the screen. This should not happen.

setnes:
I rebased the pull request.  I hope I did it correctly.  :)

Instead of the down arrow getting the original size, maybe it should call the same logic that we have under _resetZoom.  This way, the down arrow would never make an image larger than the window.  If we made that change, it's worth pointing out that the only time the up and down arrow would affect the image would be if the image was smaller than the window.  Then "up" would zoom the image to the window size and "down" would set it back to 1:1.  In the case where the image was larger, both "up" and "down" would scale the image to the window size... thus appearing to have no affect... unless the user had previously zoomed another way (mouse, etc).

I think the above behavior would be just fine as long as we provided another key that always displayed the original, 1:1, size.  The original size button (down arrow) has been very valuable to me for testing the preview sizes, and zoom logic.  :)

I assume this is all in slideshowcontrols.js?

oparoz:
Looks good :)
I'm going to test it to see if we still really only request images of certain dimensions.

Regarding the keys, I believe that what you've described is how it works now, so this PR should keep the functionality as is and we can make mods in another one.

If down brings the image back to it's starting position (smaller or fitting in the window), why would a regular user want to get the original size for cases when it's larger than the window?

Looking at the code, you simply need to modify this method to make it behave like _resetZoom
https://github.com/owncloud/galleryplus/blob/master/js/slideshowzoomablepreview.js#L115-L124

The only difference between the two methods is that one "flies", while the other resets.

setnes:
I think there might be a difference between a gallery owner and a visitor to a shared gallery.  The visitor wouldn't need to know anything about the scaling.  It should just work without any knowledge of keys.  Maybe they would discover the left and right arrow keys... that's probably it though.  If you own the images and care how they are being presented, you might go hunting for other keys... read some code... read some docs, etc.

I would propose we have the 0 (zero) key continue to provide a 1:1 zoom, but change the down key to work as coded in _resetZoom described above.

Also, in looking into all of this I discovered the F key for full screen.  This currently doesn't work the same as everything else we have just fixed... so perhaps I need to look at that now.  Using F11 in the browser before starting the slide show works better than using the F key once in the slide show.

Also, what is this about ignoring the image/svg+xml mime type?  Do we need to check that before doing any scaling?

oparoz:
I still think we should modify the shortcuts in another PR. We will probably need to make more than one change. They are not documented yet, so it's not a big deal. Since the controls auto-hide, maybe we could even have a help button in the slideshow at some point, because there is no way a user can intuitively figure out which shortcut can be used for what.
The way the whole zooming works needs to be discussed, probably at the conference. Photoswipe uses a different approach, kind of like what Flickr is doing, by only switching between the fit to screen and original sizes. You control everything with the mouse, trackpad or touchscreen.

So, for now, it should work like before and there is no zooming taking place if the original picture is larger than the screen.

What's the problem with full screen? Is it asking for a different preview size once in fullscren mode? It worked fine here as I was able to zoom in/out. Pictures looked blurry, but the originals are usually smaller.

SVG is a vector format, so there is no original size, they can fit the screen without any quality degradation.

setnes:
The problems with fullscreen mode...

1)  It is possible for the requested image to be larger than the screen.  When this happens, it is displayed 1:1 because the call to _resetZoom() is specifically skipped when fullScreen === null.

2)  The user can actually get better images by putting the browser into fullscreen mode with an F11 BEFORE entering the slide show than if the F key is hit during a slide show.  The reason is the preview image size is computed when the slideshow starts, not before each image.  So if the browser is put into fullscreen before the slideshow starts, the preview images are sized correctly for the screen.

3)  We're not consistent between fullscreen mode and regular mode with the image scaling.  Fullscreen will always zoom a smaller image to fill the screen.  Regular mode will never zoom (automatically) beyond 1:1.

My thoughts about these issues:  

I think number 1 could be fixed in this pull request if we need to keep the fullscreen mode.

I think number 2 actually makes a strong case for eliminating the special fullscreen key and logic.  F11 is pretty much a browser standard.  We could just leave this up to the user to do and remove the key/logic.  (simplification is sometimes good) 

I'm guessing number 3 was probably by design... I just don't like enlarging images beyond 1:1.  :)

As far as SVG... Sometimes we call _resetZoom for SVG, and sometimes we don't.

(this post is too long... sorry)

oparoz:
1) OK, I've finally been able to see this.

2) Indeed and this can be easily fixed in a PR. I'll wait for this one to be merged so that you don't have to rebase.

3) Correct. This is by design, but maybe it needs to change. The idea is that if you're asking for full screen, you really want the picture to fill the screen and not end up with a small image filling half of it.

Regarding the F key, not all devices have a F11 key, so it's a convenient way to have full screen mode on those, but I agree that it could be removed if we replace it with a button.

For SVGs, I still wanted to be able to get to the original size, but they're allowed to fill the screen, but it could be that I missed something.

I don't mind long posts. It helps document features :)

setnes:
I think I have the zooming the way you want it for now.  There is no zooming (in or out) if the image is larger than the window.  I also fixed the fullscreen issue such that images larger than the window are scaled to fit.  Fullscreen still works as it did before with smaller images (zooms to fit).  I moved some of the IF logic into _resetZoom and made it so startBigshot will always call _resetZoom.  I hope that's OK.

I'm not impressed with how SVGs are working yet.  I tried both zooming and not zooming on some samples.  Not all SVGs behaved the same.  One, in particular, actually broke the gallery view when I had it in the folder.  It caused everything to be on one row.  I'm guessing it may have had NULL dimensions.  I smell a new issue coming for it.

Example SVGs.
This zooms great:  http://dev.w3.org/SVG/tools/svgweb/samples/svg-files/AJ_Digital_Camera.svg
This breaks bad:  http://dev.w3.org/SVG/tools/svgweb/samples/svg-files/android.svg

oparoz:
I've just tested and in full screen mode, I can't got back to the original size for normal sized images, only for very small ones.

I didn't experience a broken gallery with the Android logo, but it doesn't zoom properly in the slideshow with the change you've made. I think it's because you've moved some of the logic.

setnes:
I fixed the full screen zoom issue you found.  I'm still looking at the SVG logic related to zooms.

The android logo very much causes the row length calculation to go screw balls in the thumbnail view for me.  If it's close to the end of the list of photos, it doesn't hurt anything.  Try with a couple rows of photos and then re-sort until it's towards the beginning of the list.

oparoz:
Good job :)
Regarding the SVG logic, I think the svg condition in resetzoom has to be split in two.

I've been able to reproduce the problem with the Android logo :). That will be for another PR.

setnes:
I fixed the SVG logic by preventing SVG images from being treated as small images.  This was a long debug session to figure out that many of the SVG images I was testing were treated as small images and therefore not zooming properly to fit the window.  The android logo was one of them.

Don't worry, I'm not touching the SVG thumbnail issue in this pull.  :)  That was an existing problem.

oparoz:
I've just tested the latest commit and everything seems to be fixed, which is great :)
You'll be our resident expert on the zooming feature :)

Now do you know how to squash commits? It's very easy if you don't.

setnes:
I did not know how to squash.  As usual... I hope I did that correctly.  :)

setnes:
I think fixing #166 here would be trivial.  Instead of $(window).width() and $(window).height(), we could use screen.width and screen.height when computing the longEdge.

oparoz:
Excellent job! Sorry there was more work involved than initially planned.

I'm ready to merge this, but since you've not signed the ownCloud contributor agreement (you still can), I need you to confirm that your contribution is made under the MIT license. [More info](https://owncloud.org/contribute/agreement/)

oparoz:
And you're correct regarding #166 :)
You can fix this as part of this PR if you want. No need to squash afterwards. Just add "Fixes #166" to the commit comment (2nd line) as it auto-closes the issue when merged.

setnes:
The fix was trivial, but I'm starting to worry about the preview file sizes.  Do the previews always have to be PNG?

oparoz:
> The fix was trivial, but I'm starting to worry about the preview file sizes. Do the previews always have to be PNG?

That's the way it's been designed, but it could change in the future https://github.com/owncloud/core/issues/12008

setnes:
Good to know it might change.  I agree with your proposal to have jpg, raw, and tiff get jpg previews and everything else to get png.  That seems like a good balance.  I have also read about some lossy ways to pre-process png so they are smaller.  https://pngquant.org/  Supporting jpg would likely be way easier though.

oparoz:
Pre-processing would slow things down too much and add new requirements to ownCloud, so that will probably never happen.

Have you given some thoughts on the way you want to license your code?

setnes:
I somehow missed your post about the license agreement yesterday.  I'll look at that in a bit.

setnes:
If I understand this correctly, signing the ownCloud agreement is required for commits to ownCloud core, but not commits to apps.  Signing the agreement will require I get my employer involved.  This might not be a fast process.  The loophole for app contribution appears to be just to state the MIT license for the code fix.

There's nothing really new with these changes.  They're basically just a refactoring of how the existing code is calling the included Bigshot image libraries.  The MIT license makes sense here.

oparoz:
That's correct. It's not required for app. but it's not clear to me whether apps mean 3rd party apps distributed outside of ownCloud or all apps.
Since Gallery is/will be the only app included with ownCloud Enterprise, I'm just being careful.

I don't think you need to get your employer involved, unless he owns code produced by you.

I started by contributing under the MIT license, until I decided to submit larger improvements to core.

So I'm going to merge this now and let's see what your next contribution is going to be and if you need to start becoming an ownCloud "collaborator" :)

Issue #160, opened 2015-05-25
Slideshow: should the browser back button behave the same as the X icon? [$5]


setnes:
The browser's back button doesn't behave as expected.  If clicking the big X in the upper right side of the screen exits the slide show, perhaps the back button should do the same thing.  I feel a bit trapped in the slide show on a mobile device if the back button doesn't actually exit.  I'm not sure what is correct here from a UI standpoint though.  One could also make a case that back should just go to the previous image.
### Files

How it works
- Folder1->Folder2.
- Slideshow is launched. Folder2 is renamed Folder2#slideshow, but it's the same position in the history
- Pressing close rewrites the URL again, back to Folder2, but pressing back goes back to Folder1!

Pressing on "close":
- keeps the position in the files list
- "erases" the slideshow from history by rewriting the URL

Pressing on "back":
- gives the impression of going two steps back since the slideshow doesn't modify the history.
-  loses both the position in the list and the folder from which the slideshow was launched
### Files
- [x] Exact vertical position in Files is not lost when exiting the slideshow via the back button
- [ ] Files app is not reloaded when exiting the slideshow via the back button. **Blocked by owncloud/core#14828**
- [x] Pressing back several times goes through the entire history
- [ ] Pressing forward after exiting the slideshow reloads the last slide which was in view in the slideshow. **Blocked by owncloud/core#14828**
### Photowall
- [x] Exact vertical position in Photowall is not lost when exiting the slideshow via the back button
- [x] Current album is not reloaded when exiting the slideshow via the back button
- [x] Reloading the app while browsing the slideshow should load the exact same slide
- [x] Pressing back several times goes through the entire history
- [x] Pressing forward after exiting the slideshow reloads the last slide which was in view in the slideshow
### General
- [x] Everything above works both in the private and public spaces
##

setnes:
I did a few more tests.  It looks like the browser history is handled differently depending on if we start in the files view or the gallery view. 

If I... 
1. start a slideshow from the **files** view
2. flip through photos
3. click the browser back button 

...I am taken back to the gallery thumbnail view.  It's not the files view, but at least the back button takes me out of the slideshow.

If I...
1. start a slideshow from the **gallery** view
2. flip through photos
3. click the browser back button 

...the URL changes to end with !.  Another back button click changes to the URL to the previous photo, but nothing changes on the screen.  If I keep clicking the back button, I can watch the URL change through photos (without loading them) until eventually I back out to the gallery thumbnail view.

[...]

PR #177, opened 2015-06-21
Stop the slideshow when using the back button.

setnes:
Pull for issue #160.


oparoz:
I'll close this and credit you in my pull request since I've moved that method to the Slideshow.

PR #191, opened 2015-06-26
fix image titles / labels

jancborchardt:
First one in a row of promised fixes to Gallery+ @oparoz :)
- unbold the image name – we only use bold when it’s important
- fix the ellipsis actually appearing by setting width to 100%
- making the title appear faster, 250ms (old) s 50ms (new), just like the user and apps menus – can I use OC.menuSpeed here like that or is that somehow private / whatnot?
- for some reason there was a `>` before every filename. That just looked weird so I removed it.
- use the same background as for album names (.5 alpha)
- also center the image titles, same as for album titles

Hope that’s cool @oparoz, please review and @owncloud/designers as well!


jancborchardt:
And another thing, something I wanted to do since a long time: Use a gradient for the background of the album & image titles. That way the titles blend in much nicer! :) cc @owncloud/designers please review

oparoz:
OK, so here are my thoughts
- Good call on the ellipsis
- I don't think the combination of the regular font + shading works for pictures with a light background. This could change in the future if we implement https://github.com/owncloud/galleryplus/issues/84, but right now it's a bit of a hit and miss
- Regarding the speed, I think this is a matter of preference. This is not a menu. I find a slower speed more elegant, but maybe we could use something in the middle?
- The `>` was an artistic decision ;), but I'm fine with its removal
- OK for centring the titles, even though they will probably end up on the left again once "likes" and comments are added on the right

Also, please use the dev branch when building your PRs :). The ultimate branches are just for people to see what the Gallery would look like with all PRs merged. No need to do anything for this one. I've created a PR against master and I'll cherry pick your commits before merging.

jancborchardt:
@oparoz I added a commit to fix the shading & added a slight text-shadow to the text (only an outline) to make it more readable. I think like that it works nicely.

Everything cool like that then? I would definitely like the animation to be fast like that. 250ms is waaay too slow and makes the app feel sluggish.

oparoz:
@jancborchardt - OK, I can always add a setting later for people who want a different animation speed, but I agree that if you want to give the impression that the app is super responsive, it helps to have a low value.

Regarding the labels themselves, there is definitely an improvement, but it's still a bit problematic.

new_title_on_light_bkg

Can you do better? Maybe just like for the background and slideshow UI, it will be tricky to have a 100% working solution until https://github.com/owncloud/galleryplus/issues/84 is implemented.

oparoz:
Thinking about it some more, even if we detect if the picture is light or dark, it might look strange if we switch label colours when rolling over different images

jancborchardt:
That’s indeed really strange-looking! The text seems to be a bit transparent as well, that shouldn’t be. Which browser do you have?

Let me investigate that more. Can you give me the PNG of that image so I can test it here?

oparoz:
Both the original and the preview are available in the demo
https://oc8demo.interfacloud.com/index.php/apps/galleryplus/s/4jjwPv5E21l8Lnk#Illustrator

And it does look better on Chrome. I was using Firefox.

jancborchardt:
@oparoz can we merge this for now? Will fine-tune this more, but atm I don’t want to pile up more into this pull request because the history already looks quite weird and I also want to work on other aspects of the app.

oparoz:
No problem

oparoz:
I'm closing this. The merge will happen here:
https://github.com/owncloud/galleryplus/pull/192

And will then be pushed to all the other branches

PR #192, opened 2015-06-27
Fix images Backport of #191

oparoz:


jancborchardt:
(Replying here) Ok, cool! So from now on I just also issue PRs to master?

oparoz:
If it's convenient for you, then yes. If you're on 8.1, you can always create PRs against dev, but it would be best to ask me to rebase before you start and there may be times, like now, where it's best to wait for some PRs in master to be merged before starting anything on dev.

jancborchardt:
Yeah, master is then most convenient. I really hope the model changes to the one we use in core and other apps too, especially because this is also confusing for new folks. ;)

oparoz:
It's confusing for core devs, but imagine if master was on 8.2 right now. Only core devs have it installed. Some contributors have probably switched to 8.1 because it's almost been released, but most hobbyists will be on 8.0 and I'm almost certain none of them will want to set up 8.2 to be able to submit patches for that, hoping for them to be backported to 8.0.

I think once this is official, I won't have a choice. It needs to me maintainable long term by anybody, but I'm not sure we'll get many contributions by people trying to fix problems they're currently experiencing with their cloud.

jancborchardt:
Yeah, I mean once it’s the official Gallery app (hopefully with 8.2) then it’s really best to align the release schedule (and backport handling etc) with ownCloud’s versions itself. Maybe @DeepDiver1975 or @Raydiation also have comments about that, but this is more a discussion for #owncloud-dev rather than this issue. :)

oparoz:
Yeah, that discussion is already spread over 3 issues/PRs :D

jancborchardt:
Thanks! Will get to more improvements soon. :)

PR #207, opened 2015-07-05
Slideshow cleanup

### LINKED ISSUE #160: Slideshow: should the browser back button behave the same as the X icon? [$5]

oparoz:
- Back button support. Fixes #160 with the help of @setnes
- Clean up all images when closed
- Remove duplicate events
- Properly clear previous/next image
- Wait for the slideshow to be ready before trying to use it (before that it was just luck)
- Load the slideshow only when needed on the Files app
- Load the slideshow at init time in the Gallery app
- Minor fixes and JS optimisations

**This is for the most recent release of ownCloud: 8.0**

Visually, you should notice that there shouldn't be any picture overlap any more when quickly switching to the previous slide before the current one has loaded, but I doubt you'll see anything else.

Using the slideshow's close button or the browser's back button should achieve the same thing on both the Gallery and Files app. Full specs are in #160

@setnes @vader2014 @demattin @jancborchardt @PVince81 

Feel free to let me know if the JS can be better written, etc.

### DIFF
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -168,3 +168,8 @@
 	filter: alpha(opacity=90);
 	opacity: .9;
 }
+
+#slideshow > .bigshotContainer {
+	width: 100%;
+	height: 100%;
+}
--- a/js/app.js
+++ b/js/app.js
@@ -1,5 +1,6 @@
 /* global OC, $, _, Gallery */
 $(document).ready(function () {
+	"use strict";
 	Gallery.hideSearch();
 	Gallery.utility = new Gallery.Utility();
 	Gallery.view = new Gallery.View();
@@ -17,7 +18,6 @@ $(document).ready(function () {
 		// Needed to centre the spinner in some browsers
 		Gallery.resetContentHeight();
 		Gallery.showLoading();
-
 		$.getJSON(Gallery.utility.buildGalleryUrl('config', '', {}))
 			.then(function (config) {
 				Gallery.config = new Gallery.Config(config);
@@ -76,6 +76,7 @@ $(document).ready(function () {
 });
 
 window.onhashchange = function () {
+	"use strict";
 	// The hash location is ALWAYS encoded
 	var path = decodeURIComponent(window.location.href.split('#')[1] || '');
 	var albumPath = OC.dirname(path);
--- a/js/app.js
+++ b/js/app.js
@@ -1,4 +1,4 @@
-/* global OC, $, _, Gallery */
+/* global OC, $, _, Gallery, SlideShow */
 $(document).ready(function () {
 	"use strict";
 	Gallery.hideSearch();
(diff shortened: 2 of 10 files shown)


oparoz:
The first commit has tmi, so here is what was done
- Added a container for bigshot/images, so that it can be more easily cleaned up as divs were not removed when closing the slideshow https://github.com/owncloud/galleryplus/commit/9986ae98d1dee1c44a50df6b3574283c2a0a6550#diff-fa1aee2a2b0036a8340cbf89a295249aR12 https://github.com/owncloud/galleryplus/commit/9986ae98d1dee1c44a50df6b3574283c2a0a6550#diff-cf40a03ac29e6718ec10b447dcabcc31R13
- The close button in the slideshow now goes back one step in history https://github.com/owncloud/galleryplus/commit/9986ae98d1dee1c44a50df6b3574283c2a0a6550#diff-5e4ddda58d290ec339ec9e57444801ddR286
- Going one step back in history when in the slideshow, closes everything https://github.com/owncloud/galleryplus/commit/9986ae98d1dee1c44a50df6b3574283c2a0a6550#diff-da2846a114690f2d9e41a05c414598a7R39
- Initialise the slideshow only once instead of creating/destroying it every time.

oparoz:
Forgot to notify @libasys :)

oparoz:
Can be tested by using `ultimate8` on core `stable8`

PR #264, opened 2015-08-07
Switch to Codeception for testing

oparoz:


DeepDiver1975:
I honestly question if this is the right way to go.

We dont use codeception anywhere else.
It will be just another entry barrier for new contributors.

Well -  lets See where this is leading us ...

oparoz:
> We dont use codeception anywhere else.

That's correct and it has not been missed so far, because for unit and basic integration tests, PHPUnit is enough. ownCloud has a few acceptance tests in various folders, but they require different environments, depending on who has written them.

> It will be just another entry barrier for new contributors.

I don't think it will, since Codeception understands standard PHPUnit tests and syntax.
I didn't have to modify a single test and still got my coverage report in Scrutinizer. The only thing I've changed is make Travis run `codecept` instead of `phpunit`.
For devs testing locally, it's a matter of running `codecept run unit` instead of `phpunit -c phpunit.xml`

There is a barrier of entry for people who want to contribute to the bootstrapping code for the acceptance tests, but I don't think that will be done by new contributors. They will focus on writing the test themselves.

The main advantages:
- Easy to set up. As a contributor it's as easy as just downloading the phar file
- One command to run all tests
- Easy to understand structure in tests folder
- Easy to test API/ajax endpoints. Requires 15 minutes to read about the syntax to use
- Easy to write acceptance tests, which come with screenshots. Same syntax as above, but with some new keywords
- Offers a choice between standard PHPUnit test and Codeception Cest for unit testing. That could be a problem for newcomers if most test are Cest classes, but they are easy to read and contributors can still use PHPUnit if that's easier for them.

The main problems:
- It doesn't understand Gherkin scenarios. That could be a problem for core if you have your BAs collecting requirements in that format. Regarding this project, I don't think many people will approach the maintainers with such scenarios, expecting them to implemented a feature and even if they did, it's relatively easy to translate them to tests. Also, Codeception can generate scenarios from the tests themselves, so you can still compare both the original specs and Codeception's text output to make sure they match

Let me know if you have specific concerns I may have overlooked :)

PR #291, opened 2015-09-01
[Testing] Test OCS provisioning API

oparoz:
# DO NOT MERGE

This can be re-used if the OCS provisioning API stops working again.


oparoz:
Obsolete now. The other core testing branch can be used

PR #294, opened 2015-09-02
Use .text instead of .html

LukasReschke:
@oparoz You might want to backport this to the appstore releases…

### DIFF
--- a/js/galleryimage.js
+++ b/js/galleryimage.js
@@ -95,7 +95,7 @@
 						image.path);
 
 					var imageLabel = $('<span/>').addClass('image-label');
-					var imageTitle = $('<span/>').addClass('title').html(
+					var imageTitle = $('<span/>').addClass('title').text(
 						OC.basename(image.path));
 					imageLabel.append(imageTitle);
 					a.hover(function () {


oparoz:
I'll update it, but is it just a good practice thing or is there a real risk here? I don't see how the filename can be used against the user in this case.

LukasReschke:
> I'll update it, but is it just a good practice thing or is there a real risk here? 

Just upload a file `<script>alert(1)<script>` and access the Gallery app and check the HTML. This leads to XSS.

It's not directly executed in this case since we use CSP but [for example MS IE does not support CSP](https://caniuse.com/#feat=contentsecuritypolicy).

LukasReschke:
See https://www.owasp.org/index.php/Cross-site_Scripting_(XSS) for some more explanation.

oparoz:
Yes, but I didn't think it was possible to create such a file. Linux just let me, so there is a real risk :|

LukasReschke:
> Yes, but I didn't think it was possible to create such a file. Linux just let me, so there is a real risk :|

Well just do via the web interface ;)

oparoz:
> Well just do via the web interface ;)

:-O  that worked... but that's a problem because those files won't sync.

PR #297, opened 2015-09-03
adjust app icon to fit new filetype icons

jancborchardt:
Please review @oparoz @deMattin @owncloud/designers


MorrisJobke:
:+1:

oparoz:
@jancborchardt - Is it a good thing to make it look like a file type though? Especially since it can show both bitmap and vector formats and we plan on adding movies as well.
Maybe go for something new for the launch of "Gallery"?

jancborchardt:
@oparoz yes, it should be uniform so it can be connected. As mentioned before: The distinction between if it’s vector or bitmap is not really important in the interface.

oparoz:
@jancborchardt - Yes, but when movies are added, then the metaphor won't work any more. I guess this can be dealt with later...

Issue #293, opened 2015-09-02
Routes must not be used before app.php is fully loaded


RobinMcCorkell:
After some routing changes in ownCloud 8.2, apps must be loaded before app routes are loaded. This means that app.php cannot directly use any self-referencing routes, for example with the URLGenerator. This is an issue with the navigation entry registration, which tries to create a URL pointing to the gallery app.

The solution is to move the navigation entry generation to a closure that returns the array, so that loading the routes is delayed until the navigation entries are actually used. This will also increase performance.

oparoz:
OK. Is there some reference implementation?

RobinMcCorkell:
@oparoz It is literally just wrapping the array in a `function() { return [...];}`, then passing that to `->add()` instead of the raw array.

oparoz:
Got it, thanks @Xenopathic :)

PR #298, opened 2015-09-03
Delay navigation route loading

oparoz:
- [x] Fix for #293

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -19,33 +19,37 @@ use OCP\Util;
 $app = new Application();
 $c = $app->getContainer();
 $appName = $c->query('AppName');
-$urlGenerator = $c->query('OCP\IURLGenerator');
-$l = $c->query('L10N');
 
 /**
  * Menu entry in ownCloud
  */
-$navConfig = [
-	'id'    => $appName,
-
-	// Sorting weight for the navigation. The higher the number, the higher
-	// will it be listed in the navigation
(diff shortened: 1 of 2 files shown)


oparoz:
So, the first patch is for you, @Xenopathic . Thanks for fixing the OCS routes and letting me know about the delayed routes.

oparoz:
All tests pass except the ones which require the return of the deleted icons in core.

RobinMcCorkell:
:+1: (if my +1 means anything in this repo that is :smile:)

oparoz:
Yeah, that's the thing. Almost no-one is familiar with the code, so there is little point in waiting for a +1 when I make little changes, but since this is AppFramework code, it should be understandable by other devs.

Thanks for the review :)

oparoz:
@Xenopathic - Now this is untestable... 

`$navigationManager->getAll()`

>  [Symfony\Component\Routing\Exception\RouteNotFoundException] Unable to generate a URL for the named route "gallery.page.index" as such route does not exist.

RobinMcCorkell:
That can only happen if the app routes are not being loaded. It's unrelated to this change...

oparoz:
The app routes are never loaded if you're not logged in if I'm not mistaken, so I guess I just didn't see it before because the app.php file is just a bunch of method calls, so PHPUnit was reporting all of these as covered.

Issue #287, opened 2015-08-31
Public controllers are broken


oparoz:
There are no integration/API tests for the public controllers because they're slow, but it's a problem because several routes were broken and not detected.
- [x] Fix the public controllers
- [x] Add one API test per route

PR #300, opened 2015-09-03
Fix public routes

oparoz:
Fix for #287, includes some slow, but necessary route tests


oparoz:
All tests pass except the ones which require the return of the deleted icons in core.

Issue #296, opened 2015-09-03
Error page should not display user controllable message if possible


LukasReschke:
`/index.php/apps/gallery/error` should not show an user-controllable error message if possible, the reason behind this is that users usually trust the response returned by a server. This is also known as [Content Spoofing](https://www.owasp.org/index.php/Content_Spoofing).

Example: `http://localhost/master/index.php/apps/gallery/error?message=Please%20send%20your%20password%20to%20owncloud%40evil.de&code=500`

![2015-09-03_14-03-54](https://cloud.githubusercontent.com/assets/878997/9657810/a204d98e-5244-11e5-8eec-81280a34558e.png)

oparoz:
So just remove the error message altogether and let the user sort it out by looking at logs?

LukasReschke:
> So just remove the error message altogether and let the user sort it out by looking at logs?

Or you use ICrypto to generate an encrypted error message and decrypt it :-)

oparoz:
OK

oparoz:
Should I put the password in the session?

LukasReschke:
> oparoz added the ready label 9 hours ago

Can you explain what the "ready" label means? :-)

> Should I put the password in the session?

Fair point. An secret per session might make sense. But then we're also at a point where we could just put the whole error in the session and read it from there and once it is displayed clean it out of there?

oparoz:
> Can you explain what the "ready" label means? :-)

Kanban: Ready, In progress, On hold
Ready issues have been accepted and will be worked on next

> But then we're also at a point where we could just put the whole error in the session and read it from there and once it is displayed clean it out of there?

Yes, that should work. Send to template and remove from session.

oparoz:
Fix is in https://github.com/owncloud/gallery/pull/301

PR #301, opened 2015-09-04
Keep error message in the session

oparoz:
Fix for https://github.com/owncloud/gallery/issues/296

@LukasReschke - Let me know if that meets your requirements.

The key points:
- https://github.com/owncloud/gallery/compare/keep-error-message-in-the-session?expand=1#diff-a8d3ad1f02ffca9503fa85aff518fc1dR143
- https://github.com/owncloud/gallery/compare/keep-error-message-in-the-session?expand=1#diff-e1b15e88afed87f9c3281c683fd9bab8R123
- https://github.com/owncloud/gallery/compare/keep-error-message-in-the-session?expand=1#diff-4b12e90075cad7aa65433b7c4bfc8063R110

### DIFF
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -78,7 +78,8 @@ class Application extends App {
 				$c->query('Request'),
 				$c->query('Environment'),
 				$c->query('OCP\IURLGenerator'),
-				$c->query('OCP\IConfig')
+				$c->query('OCP\IConfig'),
+				$c->query('Session')
 			);
 		}
 		);
@@ -125,6 +126,7 @@ class Application extends App {
 				$c->query('ConfigService'),
 				$c->query('SearchMediaService'),
 				$c->query('DownloadService'),
+				$c->query('Session'),
 				$c->query('Logger')
 			);
 		}
@@ -139,6 +141,7 @@ class Application extends App {
 				$c->query('ConfigService'),
 				$c->query('SearchMediaService'),
 				$c->query('DownloadService'),
+				$c->query('Session'),
 				$c->query('Logger')
 			);
 		}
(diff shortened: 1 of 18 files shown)


oparoz:
Rebased

oparoz:
Rebased

oparoz:
@karlitschek @DeepDiver1975 Needs to be reviewed today. Feel free to delegate to someone who can.

karlitschek:
@LukasReschke @PVince81 @schiesbn @nickvergessen can you help with a review?

LukasReschke:
Works :+1:

On a related note it might make sense to move this to the cookie for better UX but given that this only is used in the download controller this should be good as is anyways.

oparoz:
Thanks for the review.
Just to be clear (for me). You mean that retrieving the info from the cookie is better than going: session->template->browser?

LukasReschke:
> Just to be clear (for me). You mean that retrieving the info from the cookie is better than going: session->template->browser?

Yes. PHP sessions are blocking unless closed (which we do per default). So if a user clicks download and you set the `UseSession` annotation he can't do other things within the same session until the request is completed (other sessions will still work fine). Such as going to another picture and downloading this as well.

Though this is a pretty minor use case.

oparoz:
Thanks.
I'd like to get rid of the customised error message for downloads. Session or cookies should not be used in a RESTful API anyway (and I want api controllers to mirror the standard one to keep things simple), but I need to make sure it still works with encryption. All the public controllers need `UseSession` in order to have access to the encrypted files.

oparoz:
I've removed the session from the download method and it still works with encryption.

Tests fail because `core` master is broken. They all pass locally, when not running in single user mode.

PR #303, opened 2015-09-06
More tests for preview controller

oparoz:


oparoz:
Thanks `core` for breaking everything once more...

PR #311, opened 2015-09-08
make gallery/filelist switcher button less present and distracting

jancborchardt:
Removed the button style and made it a bit more transparent. As we talked about @oparoz @karlitschek @owncloud/designers 

I tried using the pictures filetype icon instead of the current icon but it looks wrong, an image icon floating there.

### DIFF
--- a/css/gallerybutton.css
+++ b/css/gallerybutton.css
@@ -1,14 +1,15 @@
 /* toggle for opening shared picture view as file list */
 #openAsFileListButton {
 	float: right;
-	margin-right: 4px;
 	margin-top: 5px;
 	font-weight: normal;
+	background-color: transparent;
+	border: none;
 }
 
 #openAsFileListButton img {
 	vertical-align: text-top;
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
-	filter: alpha(opacity=50);
-	opacity: .5;
+	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=30)";
+	filter: alpha(opacity=30);
+	opacity: .3;
 }


MorrisJobke:
Tested and looks good :+1:

karlitschek:
:+1:

jancborchardt:
@oparoz I’d like a final review from you here. :)

oparoz:
Hmmm... I'll have to check visually, but that bit of code is about the icon which is shown in Gallery, not in the Files app.

jancborchardt:
@oparoz it’s for both. ;) I tested it and it worked. And actually I like that it works for both buttons.

oparoz:
:-1: 
Does not work with the other buttons any more... You need to change all or none of them.
I doubt making them all transparent is going to work.

jancborchardt:
This is just for the toggle button, not for the sorting buttons.

If you add a class to it to make it just work in the files view and not in the Gallery view that’s fine too.

oparoz:
Maybe it would be best to use this as a base:
https://github.com/owncloud/gallery/commit/894d3ad507679ec9fd264a74944ac9996b8e78bc

only removing the switch for files/gallery

oparoz:
Actually, the best way forward would probably be to finish https://github.com/owncloud/gallery/pull/148

jancborchardt:
@oparoz it seems that #148 tackles a bit too many issues – I’d suggest we rather split this up? And this is the first definitive thing which we should fix. Ok? :)

oparoz:
The first commit in #148 is actually what you need in order to be able to easily work on all buttons at the same time, the other commits are actually tracked in separate issues, so will be implemented anyway, so what's left would be to move share to the left and remove the switcher classes (on top of my head) and use just the one button. When that's done, we can merge and then you can work on the design of the buttons.

The result is going to be the same as if we do 5 PRs. It could even be merged as is, with the switcher and then you can create a PR to get rid of it.
What do you think?

jancborchardt:
@oparoz I need to merge this as the release freeze has priority. We can adjust stuff in #148 after this, but it’s freeze today for core – hence things which mess with Files need to be finished.

oparoz:
> but it’s freeze today for core – hence things which mess with Files need to be finished.

I'm still not clear on that. Is it only `core` which gets the feature freeze then?

jancborchardt:
@oparoz yeah, although the main shipped apps (like Activity, Sharing, and yes, Gallery) are included in that feature freeze for stability reasons. We can do adjustments and fix bugs (e.g. like moving the download button, fix the buttons in the controls bar) but should not introduce new features (like a new fullscreen button for example).

karlitschek:
Yes. Feature freeze has to happen in all repos with apps that are shipped in the release. Future feature development has to happen in feature branches for now. And soon we will do stable8.2 branch so that master is open for features again. I hope this works for you

oparoz:
@karlitschek - What about security issues? Does it matter if it's labelled an enhancement -> https://github.com/owncloud/gallery/issues/296 or has no label -> https://github.com/owncloud/gallery/issues/347 ?

I'm missing a wiki entry on this btw.

oparoz:
I'm also confused as to why this slight alteration of button design had to make it by the feature freeze deadline @jancborchardt . It's not a new feature, just a design fix that could have waited a bit, no?

jancborchardt:
@oparoz as I said: Since it inserts something new in the Files app, that is, a new feature which is _very_ visible, it should be polished for the feature freeze already. Also, the PRs for it were laying around for long enough. ; )

Design tweaks like this in the Gallery app itself still have some time (though better earlier than later!) but for things which touch Files, I wanted to have it clean for feature freeze already.

oparoz:
@jancborchardt - I now understand the rationale, but it didn't introduce a new feature, the "Pictures" button has been there for a very long time; at least since 7.0 :
https://github.com/owncloud/gallery/blob/stable7/js/public.js#L18-L27

I'll probably introduce a new label to clarify what needs to go in by the deadline as it seems to me that core still hasn't fixed all the design issues, even though I saw you go through lots of open issues last night :)

LukasReschke:
> @karlitschek - What about security issues? Does it matter if it's labelled an enhancement -> #296 or has no label -> #347 ?

Security changes which technically are regressions since it behaved differently before are fine to be merged past feature freeze. As long as we don't add new security features fixes with regard to this are okay.

oparoz:
Thanks @LukasReschke

jancborchardt:
> the "Pictures" button has been there for a very long time; at least since 7.0

@oparoz in Gallery+ maybe, yes. But this is definitely the first release where the view switcher is directly in the controls bar of Files, injected by the shipped Gallery app.

oparoz:
No, no, load stable7 and see for yourself ;)

It's just inconsistent and appears/hides depending on if you have JPGs in the folder or not.

jancborchardt:
Ok, then I maybe missed that. :\ Still it’s better to have prominent stuff like this polished as early as possible.

Issue #226, opened 2015-07-22
Make background colour switcher optional


jancborchardt:
The background color switcher for transparent images is really not necessary. We should do it like Wikipedia and just insert the white-grey pattern background: https://en.wikipedia.org/wiki/Wikipedia#/media/File:Wikipedia-logo-v2.svg

oparoz:
Unfortunately, that doesn't work with white content.

jancborchardt:
This is how it looks in the Wikipedia viewer: https://commons.wikimedia.org/wiki/Category:White_logos?uselang=de#/media/File:Disqus_logo_%28white%29.svg
Indeed not ideal, but if we just increase the contrast of the white and grey, it will be nicely visible.

And even then, I would say if anything a background switcher like that should be a shortcut or something else, but not a prominent button like it is now. Because it really is an edge-case.

oparoz:
We've had this discussion before :). Some people have folders filled with such content and the slideshow becomes useless if they can't present it in a decent way.

I thought about making the checkboard darker, but then things start looking strange, as it can interfere with darker images.

I'll just make the feature optional, via the config file first.

deMattin:
Example Preview (example png with mixed transparent and solid)
with white background:
![image](https://cloud.githubusercontent.com/assets/5565503/8830319/eb9652dc-309b-11e5-9969-a4183844ac1f.png)
with black background:
![image](https://cloud.githubusercontent.com/assets/5565503/8830327/f4b3cc8c-309b-11e5-9890-c692a8dfe5b2.png)
This is, how it would look like with a pattern (simulated with photoshop, that uses this to show transparency):
![image](https://cloud.githubusercontent.com/assets/5565503/8830649/aa290464-309d-11e5-9234-3d70b5357030.png)

The problem is, that you are only able to see the real looking of the transparent picture with switching.
I would even like to have a third color (grey or even colored) to circle through these backgrounds!
I like this background switching and when designing transparent graphics, I think, most people use different colored layers in working progress to check the looking of transparency.

With background switching, I am able to show you a graphic and you are able to assess it.
With a white/grey or single color background it's a matter of chance, if the picture is shown in an appropriate manner.

All in all: If you don't need this button, you are not forced to press it!
But if transparency isn't shown appropriate, you have to download the file and open it in a fitting graphics programm and preview is useless in his case.

Edit:
In above example, you see by switching, that "MK" isn't transparent as it should be!
[...]

Issue #84, opened 2015-03-11
Automatically select an appropriate background colour based on the image's luminance


oparoz:
### Current situation
- The large previews use a black viewport
- Each picture has a 30 pixels wide white border and a white background
- A toggle-background button was introduced so that the background+border could be turned off. There is a discussion going about what should be the default behaviour. See #40 
### The problem

This is not ideal as currently users have to manually adjust the background+border if they're viewing very bright pictures and the white border doesn't look very nice with dark pictures with no transparency.

It gets worse in slideshow mode as you have to adjust the background if reaching a picture which can't be properly seen with the currently selected background.
### Solution
- Calculate the luminance of each picture and adjust the background accordingly
- Keep the toggle button as an emergency measure?
- Possible bonus colour transition if #83 is implemented

**IE8 is not supported**
#### Feasibility study

I did some experiments both in PHP and JS and it's doable. 
Background switcher should stay imho, at least as an option.
- [x] Build proof of concept
#### Development

The amount of work required is estimated at 2 days, including extensive testing
- [ ]  Secure funding
- [ ]  Build specifications
#### Implementation details
- Ignore some media types, either to speed things up or because we can't easily calculate their luminance without using additional libraries
- Calculate the luminance
- Calculate the alpha level
- Calculate what the optimum background colour should be
#### QA

If the accuracy level reaches 90%, the toggle button (#40) could be removed
- [ ] Gather a large sample of very different pictures to test with
- [ ] Optimise the algorithm so that it gets it right >75% of the time. % tbd based on funding.

@anarcat - Do you have some sample pictures? Would auto white/black be OK?
@jancborchardt - This would solve the white border issues when going fullscreen and should improve usability
##

anarcat:
here's hopefully a sample: ![image](https://cloud.githubusercontent.com/assets/796623/6609679/1d3c0144-c82b-11e4-9af0-2cc7e8268c66.png)

a good test is the images generated by nagios's status map.

i still believe just having a gray background would be a good compromise.

oparoz:
Thanks! That's detected as a very dark picture and a white background is automatically selected when in full screen.

jancborchardt:
@oparoz I don’t know about this. The dark background is simply standard for image viewers. It would be incredibly janky if the background would change when switching images. This seems like overthinking it.

We shouldn’t have a background switcher, nor have it switch automatically. The only thing we wanted to solve is the issue for transparent images, and we kinda did that.

[...]

Issue #40, opened 2015-02-24
Add a button to toggle the colour of the background in the slideshow


oparoz:
As described here by @anarcat 
https://github.com/owncloud/core/issues/11418#issuecomment-58079695

Happens as soon as you're starting to preview more than just JPEGs.
### Requirements
- [x] Design a toggle button. I don't see which standard oC icon we can use. Suggestions welcome.
- [x] Place the button next to the download button
- [ ] Make the controls work with a light background. A solution could be to create a light theme https://github.com/interfasys/galleryplus/issues/51
- [x] Implement the CSS background colour decoder and the switching logic
### Open discussion
- Should the user-chosen "theme" stay for as long as the app remains running? The colour of the slideshow won't change until the button is pressed again.
- Should the theme be reset at every picture change?
- Should the theme be reset every time the slideshow is launched?

Your input (and help) is welcome @jancborchardt, @anarcat

anarcat:
Hmm... i am not sure that will solve the problem i described in #11418.

i believe it would be best to use a gray background like is done in the `eog` image viewer. it has worked well for me in general, as it is pretty rare that gray lines are drawn upon a transparent background. in such a case, maybe a background toggle would be useful, but it shouldn't flip to black either, maybe a darker gray, that's it.

my understanding of the situation is that the image rendering is unreliable: it doesn't yield consistent results. Sometimes the background is black, sometimes it's white, sometimes it's checkered. this needs to be made more consistent and reliable.

jancborchardt:
Yeah, this seems an over-the-top solution. We don’t need a »theming« toggle here.

@oparoz maybe have white as the background color for every image (only the image dimensions, not the whole background!) so that images with transparency will have white background?

Granted, this still doesn’t work for transparent images with light elements, but arguably those are more rare – and then the checkerboard pattern would still be shown on hover.

cc @owncloud/designers @icewind1991

oparoz:
@anarcat - This is to solve the problem highlighted in your comment, for full screen previews. Image rendering works, transparency is respected, but with a black background, you can't see anything when the picture is dark.
In my tests, I toggle between dark grey and black. Anything lighter, like the background in eog won't work with the current controls and that's the reason I want a light theme.
@jancborchardt - True, a picture sized light background could help. I'm going to see if it's compatible with Bigshot. That may re-introduce the checkboard pattern as well.

oparoz:
[...]

PR #312, opened 2015-09-08
remove background toggle as it was very confusing for people

jancborchardt:
Remove the background toggle button since it came up in many user tests it was very confusing. cc @karlitschek @tomneedham @oparoz 

Instead, all images should have white background by default – without border! That way we will take care of nearly all transparent pngs/svgs. We can gradually improve to also fix the corner case of transparent images with _white_ foreground.

Please review @owncloud/designers @oparoz

### DIFF
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -137,11 +137,6 @@
 	top: 6px;
 }
 
-#slideshow > .changeBackground {
-	right: 135px;
-	top: 6px;
-}
-
 #slideshow > .notification {
 	margin: 0 auto;
 	max-width: 60%;
--- a/js/slideshow.js
+++ b/js/slideshow.js
@@ -97,7 +97,6 @@
 			var currentImageId = index;
 			return this.loadImage(this.images[index]).then(function (img) {
 				this.container.css('background-position', '-10000px 0');
-				this.container.find('.changeBackground').show();
 
 				// check if we moved along while we were loading
 				if (currentImageId === index) {
@@ -113,8 +112,6 @@
 					img.setAttribute('alt', image.name);
 					$(img).css('position', 'absolute');
 					$(img).css('background-color', backgroundColour);
-					var $border = 30 / window.devicePixelRatio;
-					$(img).css('outline', $border + 'px solid ' + backgroundColour);
 
 					// We cannot use nice things on IE8
 					if ($('html').is('.ie8')) {
@@ -214,15 +211,12 @@
 			var container = this.zoomablePreviewContainer.children('img');
 			var rgb = container.css('background-color').match(/\d+/g);
 			var hex = "#" + toHex(rgb[0]) + toHex(rgb[1]) + toHex(rgb[2]);
-			var $border = 30 / window.devicePixelRatio;
 
 			// Grey #363636
 			if (hex === "#000000") {
 				container.css('background-color', '#FFF');
-				container.css('outline', $border + 'px solid #FFF');
 			} else {
 				container.css('background-color', '#000');
-				container.css('outline', $border + 'px solid #000');
 			}
 		},
 
@@ -239,7 +233,6 @@
 			}
 			this.container.find('.notification').html(message);
 			this.container.find('.notification').show();
-			this.container.find('.changeBackground').hide();
 		},
 
 		/**
@@ -364,11 +357,6 @@
 						{
 							el: '.downloadImage',
 							trans: t('gallery', 'Download'),
-							toolTip: true
-						},
-						{
-							el: '.changeBackground',
-							trans: t('gallery', 'Toggle background'),
 							toolTip: true
 						}
 					];
--- a/js/slideshowcontrols.js
+++ b/js/slideshowcontrols.js
@@ -59,9 +59,6 @@
 			// Hide prev/next and play buttons when we only have one pic
 			this.container.find('.next, .previous, .play').toggle(this.images.length > 1);
 
-			// Hide the toggle background button until we have something to show
-			this.container.find('.changeBackground').hide();
-
 			if (autoPlay) {
 				this._play();
 			}
@@ -126,7 +123,6 @@
 		 */
 		_specialButtonSetup: function (makeCallBack) {
 			this.container.children('.downloadImage').click(makeCallBack(this._getImageDownload));
-			this.container.children('.changeBackground').click(makeCallBack(this._toggleBackground));
 		},
 
 		/**
@@ -337,15 +333,6 @@
 			var downloadUrl = this.images[this.current].downloadUrl;
 
 			return this.slideshow.getImageDownload(downloadUrl);
-		},
-
-		/**
-		 * Changes the colour of the background of the image
-		 *
-		 * @private
-		 */
-		_toggleBackground: function () {
-			this.slideshow.toggleBackground();
 		}
 
 	};
--- a/templates/slideshow.php
+++ b/templates/slideshow.php
@@ -5,7 +5,6 @@
 	<input type="button" class="svg previous icon-view-previous"/>
 	<input type="button" class="svg exit icon-view-close"/>
 	<input type="button" class="svg downloadImage icon-view-download"/>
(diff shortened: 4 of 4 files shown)


karlitschek:
agreed. white background without border should work :+1:

MorrisJobke:
I like this :+1:

MorrisJobke:
> I like this :+1:

... and tested it of course :)

oparoz:
:-1: 
That's not the proper way to do this. It should remain optional for users who actually rely on this because they can't see their images in the slideshow because they have a transparent background.

oparoz:
@jancborchardt - In order to get rid of it quickly, so that you don't have to add the setting, maybe just comment out the button and I'll make it optional later.

jancborchardt:
> can't see their images in the slideshow because they have a transparent background.

As we talked about – adding a white background by default will take care of nearly all transparent images, except the ones with white foreground. This is basically the same as the current approach except it’s automatic.

And for images with white foreground it will be fixed by adding something like showing checkerboard pattern or black background on hover. Again, we talked about this. We need to move in small steps here and having a very prominent interface element is not an elegant solution here.

jospoortvliet:
I support the idea of first fixing the 95% use case before addressing the rest... And that's coming from somebody who has plenty of images which are white-on-transparent ;-)

oparoz:
The problem with merging incomplete features is that it takes forever to get the missing part done. I'd rather wait for the full PR since nothing really depends on this piece of work. On top of that there can always be unforeseen side-effects and reverting later on makes things more complicated.

I do agree that a button with no design is not elegant, but it could easily be replaced with a small colour scale letting the user pick the background he wants to use if black doesn't work. It does add some bright shades to the interface, but then no solution is perfect, apart from (https://github.com/owncloud/gallery/issues/84)
The hover solution does have problems if you're trying to go through an album containing dark designs with a transparent background (icons, gifs, etc.). You don't want to have to hover over all of them and the checkboard does not really help.

So, getting rid of the button, yes, but we need a working alternative, in this PR, unless we go for optional button or auto-luminance.

deMattin:
i don't have many white with transparent pictures because I use a colored transparency to better see, if transparency working.
But most of mixed white and transparent pictures are modificated ownclud pics!

So it would be a bit odd, to make exactly THESE owncloud core pictures completely unviewable by new official gallery.

jospoortvliet:
@oparoz I think you are letting 'perfect' stand in the way of 'good enough'...

A full background IS good enough for almost everything. The few cases where it isn't - sure, that should be fixed some day but I wouldn't lose sleep over it if it wasn't. Meanwhile, right now - we can solve it for most cases.

deMattin:
If there has to be something done NOW,without having a good solution, the checkerboard should be used as background. This way it's leastwise useable for any pictures (but not looking well in any case).
And I don't understand, why you say here, that white background fits most use cases as transparent background.
My pictures with transparency from other sources have mostly grey, white or black as non-transparent colors.
So a dark or light grey (or any color) would probably be a better choice, if you have to choose a color and no checkerboard.
**Not switchable and fixed white or black for transparency background are the worst colors that can be choosen!**

oparoz:
@jospoortvliet - The problem is that your 95% is not my 95% :). I was shocked when I first started to use ownCloud. Pictures app would only work with the Photos folder and would only show images of 3 media types. That does not work at all for SMEs, but it is perfectly fine for users dumping their phone's content to ownCloud.
All this to say, that it's usually best to fix the design rather than remove the feature, unless it's not in line with the USP, chart, etc. in which case Gallery+ can take over.

I have implemented most of the new features for a reason. There is a need for them. Not by me, but by people using the system daily and who run into those issues. In a few of those merged PRs, there is a visual design problem, no question about it (this button is simply the old file icon!). That's because that's not my trade and few people have stepped in to help me.

@deMattin - In my initial tests, I did pick a grey, the colour is in the comments, but I think it wasn't deemed to be an acceptable solution.

## Action plan
1. The data sample used for testing should be seriously augmented with real life material. 3 JPGs doesn't cut it
2. Test all/most solutions
3. Pick a solution

## Solutions

### Button to switch background

#### +
- works in 100% of the cases

#### -
- not pretty
- distracting
- needs manual intervention

### White background + hover

#### +
- works in 53.45%-95% of the cases

#### -
- Discovery issue
- needs manual intervention
- needs to keep the finger on the picture to be able to see it

### Pick the background colour from a thin ruler

#### +
- works in 100% of the cases

#### -
- needs manual intervention
- distracting

### Automatic luminance

#### +
- works in 99% of the cases

#### -
- doesn't work with SVGs, which are not supported anyway

### Make button optional + white background + hover

#### +
- works in 100% of the cases

#### -
- needs documentation so that people needing the feature can enable it
- won't be pretty for people enabling the feature ;(

oparoz:
Most of these solutions were already discussed here:
https://github.com/owncloud/gallery/issues/226

jospoortvliet:
@oparoz IF you can detect transparency, use a checkerboard for those and black for the rest... This can even be based on file type - 90% of the photo's are JPG, no transparency there - black back. GIF and PNG can have transparence - checkerboard.. ?

oparoz:
@jospoortvliet - Yes, I already pick the proper background based on the media type, but it's currently white and doesn't work for some files, so people still use the button.
Using a checkerboard pattern would be better, but it's not a silver bullet as @deMattin has shown:
https://github.com/owncloud/gallery/issues/226#issuecomment-123782127

The slideshow is a viewing tool and should work in all situations or it will lead to frustration.

Until auto-luminance can be implemented and if no decision is made before the deadline, I'll probably make the button optional and use a pretty brightness icon.
There is still the issue of the border. Without it border-to-border pictures don't look good, but that could be considered an edge case.

jancborchardt:
> IF you can detect transparency, use a checkerboard for those and black for the rest

@jospoortvliet you mean _white_ background for the rest, right? Because any normal editor like GIMP, Inkscape (yes also the Adobe ones) use white as default background.

jancborchardt:
That is, only the format of the image should have white background. The overall background of the slideshow should always be dark of course.

oparoz:
Well in bitmap editors, if you work with transparent images, you use the transparent background, which is represented by a checkboard. It's true that Inkscape uses a white background as default, but I'm sure this can be easily changed when dealing with white objects.

jancborchardt:
> It's true that Inkscape uses a white background as default, but I'm sure this can be easily changed when dealing with white objects.

It’s actually not that easy – you have to go into the settings and manually change the background. But let’s not take that as a good example.

We’re running in circles here: Default background should be white (only for the image size of course, so not visible for jpgs etc), and checkerboard on hover. This will take care of most cases. When that is done we can use that as a base for further improvements.

oparoz:
> Default background should be white (only for the image size of course, so not visible for jpgs etc), and checkerboard on hover.

OK, so please create a PR with just that. Then we can work on making the button optional and the final item on the list would be how to deal with border-to-border images (https://github.com/owncloud/gallery/issues/40#issuecomment-76074466)

jancborchardt:
@oparoz I don’t have enough JS knowledge to do that as you know. Would be cool if you can take care of it.

What would mean making the button optional? A config switch somewhere in a text file? (That would be fine by me, as with that implementation mentioned above I really think most cases will be covered …)

oparoz:
> @oparoz I don’t have enough JS knowledge to do that as you know

I'll teach you! :P , but we're talking about CSS here. A checkerboard defined in :hover if I'm not mistaken.
Actually, this should be a junior job for someone interested in the slideshow... @tomneedham @raghunayyar @setnes maybe?

> What would mean making the button optional?

Just like native SVG support is optional, you add one line to the config file and that enables the button

oparoz:
Interestingly, the new right-sidebar is having the exact same problem the slideshow has:
https://github.com/owncloud/core/issues/18796#issuecomment-139369920

oparoz:
Closing this as the button is now optional. We need a new PR for the hover.

jancborchardt:
Cool, thank you! :)

PR #317, opened 2015-09-09
[WIP] Enable encryption in tests

oparoz:
This currently does not work because there are no (encryption) hooks attached to the login method used in tests: https://github.com/owncloud/core/issues/18085

### DIFF
--- a/tests/_bootstrap.php
+++ b/tests/_bootstrap.php
@@ -23,6 +23,10 @@ OC::$loader->addValidRoot(OC::$SERVERROOT . '/tests');
 // Give access to core tests to Codeception
 Autoload::addNamespace('Test', '/../../../tests/lib');
 
+// Enable encryption
+OC_App::installApp('encryption');
+OC_App::enable('encryption');
+
 // Load all apps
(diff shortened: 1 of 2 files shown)


icewind1991:
Have a look at https://github.com/owncloud/core/blob/master/tests/lib/traits/encryptiontrait.php and https://github.com/owncloud/core/blob/master/tests/lib/connector/sabre/requesttest/encryptionuploadtest.php

oparoz:
Ah cool, thanks. I've noticed recently that core finally has traits for the things we regularly use :)

schiessle:
@oparoz what's the state of it? Would be nice to have for 8.2.1. Are you still working on it? Do you need any help?

oparoz:
I don't see this as a high priority, but something which would be nice to have in order to cover more use-cases, so I haven't looked at it since creating the PR.

Just like with some acceptance tests, I was thinking of keeping a special branch open which runs the encrypted tests. That way it doesn't take an hour to run tests. The main problem with that approach is that PRs wouldn't be automatically checked...

Any thoughts?

Issue #289, opened 2015-08-31
Remove the scrollbar when it's not required to have one


oparoz:
A scrollbar is showing up on the right, even if there are only a couple of pictures in the album. We need to figure out what triggers it and eliminate it.

oparoz:
Use master to test

ghost:
@rperezb test please?

Dianafg76:
I tested this Issue, I Have used:
1. Firefox, it is working OK. The scroollbar you can see when have more than 25 Photos, it does not appear before.
2. Chrome, it is working OK. The scroollbar can see when they have more than 25 photos , it does not appear before.
3. Safary, it is working OK. The scroollbar can see when they have more than 25 photos , it does not appear before.
4. IE8, is an incompatible browser for gallery

Closed

oparoz:
Many thanks.

Issue #199, opened 2015-06-30
Scrollbar visible in slideshow for private link shares.


setnes:
If I open a gallery folder using a private link share, the scrollbar is still visible when I start a slideshow.  If I go to the same folder when I'm logged in, the scrollbar is hidden when I start a slideshow.  This difference can only be seen if there are enough images to need a scrollbar in the gallery view.

I'm not sure if this is from a recent change, or if it's something I just never noticed before.

This does not happen if you navigate in from the file interface.

oparoz:
Thanks for the report. I did notice that on mobile when sharing links, but I never took the time to bisect the problem.
It's probably "just" a problem with the CSS. I did update the public.php template a couple of months back to keep it in line with files_sharing.

oparoz:
Seems to be a Chrome only bug...

oparoz:
Unlesss someone has a better idea, I'm going to automatically disable/enable the scrollbars when launching/quitting the slideshow.

I can "fix" body-public in the Gallery app by adding `overflow:hidden`, but I can't do the same in the core app, so the above solution is the best I could think of.

oparoz:
@setnes - Let me know if it's fixed for you in master.

setnes:
I am testing using ultimatedev and this scrollbar issue seems to be gone.

Issue #29, opened 2015-02-19
Show thumbnails as soon as they're received


oparoz:
## Current situation
- The app sends X requests to the preview system to generate Y thumbnails at a time
- At the same time, the view asks for a row
- Elements are added to the row one at a time
- The script waits for the thumbnail to be available **<- This is blocking the rendering of the row**
- Thumbnail is added to the row and total width is checked against max width (browser window)
- When the row is full, it is rendered

The thumbnails needs to be analysed so that the row can be composed to fit and look nice. If we only generate thumbnails of 400x200, then the view will be plain looking.
## Ideal situation
- Identical first steps
- As soon as a thumbnail is received, it is used to populate the row
- To the user, the view is filled one thumbnail or album at a time
#### Feasibility study
- [x] Collect requirements
- [x] Calculate effort required
- [x] Build proof of concept
- [x] Implement

**IE8 is not supported**
### Screencasts

_Note: the refresh rate in the videos is quite low, so this is not reflective of the smootheness of the animation_

**With cached thumbnails**
http://screencast-o-matic.com/watch/cDn6n7hR39

You can see that once the album is cached, there is no resizing taking place any more

**On an empty cache**
http://screencast-o-matic.com/watch/cDn6nEhR3E
### Features:
- Semi-opaque row while it's being filled
- Spinners on images in albums
- Spinners on single images on row (with caveat, see below)
### Caveats
- Since we don't have the real size of the images, the next image or its spinner are sometimes misaligned. It's currently impossible to fix and has to be accepted as a side-effect of being able to show some information before having the full picture
- When resizing the window on a desktop, there is a flashing effect as each time the wall is being redrawn, the opacity goes back to being semi-transparent until the images are resized. In my experience, the fix involves introducing and monitoring a state to know if we're currently in a resizing loop. I don't think it's worth for me to invest time to improve this, but a PR can certainly be introduced later if someone finds a better way
- On low powered devices, there may be a bit of stuttering when resizing very wide images. I find this acceptable, compared to the advantages.

oparoz:
Current setup does this
* View asks for a row
* Row asks to add the next element in the list
* Adder asks for a thumbnail before adding it to the row <- This is blocking the rendering of the row

The thumbnails needs to be analysed so that the row can be composed. If we only generate thumbnails of 400x200, then the view will be plain looking. The analyser does this:
* Did we get one?
* If yes, what is its width?

theCalcaholic:
[...]

PR #319, opened 2015-09-09
Remove unneeded scrollbar + add loading feedback

### LINKED ISSUE #130: Add a busy spinner whilst loading thumbs
The apps shows a busy spinner whilst initially loading an album. Great. 
The same should happen whilst still loading additional thumbs, probably centered in the row to come. 
Otherwise there is no means to tell if loading has finished.

oparoz:
So, while fixing #199, I may have introduced a scrollbar issue https://github.com/owncloud/gallery/issues/289, because I didn't want to be dependant on core fixing files, but there is no way around it.
@jancborchardt - Please try to trans-port the first commit to the public template in core. The div structure is not the same, so I can't just apply the patch.
Without a fix in core, the slideshow can't fill the entire viewport. There is a scrollbar on the right.

While I was there, I also fixed #130, while we wait for a better row generation engine: #29

@setnes @demattin @arkascha, please test.

Tested on Chrome and Firefox.

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -46,3 +46,10 @@ body {
 #gallery.hascontrols {
 	padding-bottom: 0;
 }
+
+#loading-indicator {
+	height: 32px;
+	width: 100%;
+	margin-top: 100px;
+	margin-bottom: 0;
+}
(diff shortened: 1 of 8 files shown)


oparoz:
Rebased

Issue #336, opened 2015-09-13
Use a higher quality spinner


oparoz:
The one we use from core simply doesn't cut it. It's very pixelated and has jagged edges.

oparoz:
Available in master. You'll notice it in the photowall, slideshow and infobox
@jancborchardt @setnes @demattin

jancborchardt:
@oparoz sorry what, did you change the spinner in the Gallery app? We should keep in line with core. We can change it to a better spinner with 9.0.

oparoz:
It looks the same, but with less problems. It can be altered slightly maybe if you think it's too big, etc.

jancborchardt:
@oparoz ah ok! Can you also change it in core then? And can you link to the change so I can check it? :)

oparoz:
Here you go: https://github.com/owncloud/gallery/commit/4ae75cb14ca8078b240c387a45b0b9ccc67169be

Fairly easy to implement once you understand how it works, but it's not as simple as adding the GIF as a background-image.

oparoz:
One think I like is that you can decide if you want to hide or remove the element. There may be advantages to each situation.

jancborchardt:
@oparoz hmm … sorry but that spinner doesn’t look good or similar to the current one – it’s too bulky.

Can you revert that change please? My plan for 9.0 is to move to a more modern and high-res spinner anyway. But for 8.2 we shouldn’t do such a strange cosmetic change only in the Gallery app.

oparoz:
No need to revert, you can make it look exactly like the old one:

https://fgnass.github.io/spin.js/

oparoz:
Default values

```
  var defaults = {
    lines: 12             // The number of lines to draw
  , length: 7             // The length of each line
  , width: 5              // The line thickness
  , radius: 10            // The radius of the inner circle
  , scale: 1.0            // Scales overall size of the spinner
  , corners: 1            // Roundness (0..1)
  , color: '#000'         // #rgb or #rrggbb
  , opacity: 1/4          // Opacity of the lines
  , rotate: 0             // Rotation offset
  , direction: 1          // 1: clockwise, -1: counterclockwise
  , speed: 1              // Rounds per second
  , trail: 100            // Afterglow percentage
  , fps: 20               // Frames per second when using setTimeout()
  , zIndex: 2e9           // Use a high z-index by default
  , className: 'spinner'  // CSS class to assign to the element
  , top: '50%'            // center vertically
[...]

Issue #334, opened 2015-09-13
SVG icons are not resized on IE


oparoz:
![tiny](https://cloud.githubusercontent.com/assets/323220/9836787/c0726e38-5a26-11e5-9e60-66ab04a086d1.png)
##

oparoz:
The solution is to use the PNGs on IE.

karlitschek:
@jancborchardt any idea?

oparoz:
It's a known issue
http://caniuse.com/#search=svg

> IE9-11 desktop & mobile don't properly scale SVG files. Adding height, width, viewBox, and CSS rules seem to be the best workaround.

One solution in order to keep SVGs for IE is to fix all the media type icons in core.

oparoz:
Keeping open, in case someone wants to play with SVG and IE.

oparoz:
In order to see the difference, you need to upload a broken JPG to a folder and to load that folder in Gallery in both Firefox and IE. One will use the SVG icon, the other the fallback.

oparoz:
This issue was moved to nextcloud/gallery#74

Issue #329, opened 2015-09-12
Override core's css which makes the share dropdown jump to the left


oparoz:
This is making the share dropdown jump to the left
https://github.com/owncloud/core/commit/8b2e09a77e22d234b48ee572b13e69001f1a3e54

We need to override it.

oparoz:
Available in master.
1. Open the share dialogue 
2. Resize the window so that it's less than 990px wide

The dialogue should stay on the right

@jancborchardt @setnes @demattin

oparoz:
It's impossible to test this right now and the share button will probably have to be removed as the logic we use has been removed from core.

oparoz:
@cmonteroluque - This will be automatically fixed as soon as #359 is fixed, which is targeted for 8.2, so I'm putting it back on 8.2

ghost:
@oparoz ok, reviewed

PR #337, opened 2015-09-13
More fixes

oparoz:
Fixes for
- SVG icons are not resized on IE #334
- Use a higher quality spinner #336
- Override core's css which makes the share dropdown jump to the left #329


karlitschek:
nice. seems to work :+1:

Issue #341, opened 2015-09-14
Wrong auto-play icon when stopping the slideshow with the spacebar


oparoz:

oparoz:
Fix is in #342 for now. Pressing the spacebar works like clicking on the play/pause icons
@setnes @demattin

jancborchardt:
Fixed by https://github.com/owncloud/gallery/pull/342

Issue #237, opened 2015-07-26
Show file name in the slideshow


oparoz:
_User requested_

Just like we can see it in the photowall, it would be great to be able to have it without having to look at the URL.

oparoz:
@jancborchardt (and co.) - To take into consideration when re-designing the interface.

jancborchardt:
Yeah, good call. We can and should combine this with a rename function directly in the slideshow. Should probably show the name centered below the image, along with the other actions like share, download etc.

jancborchardt:
This has been fixed by you, @oparoz ;)

PR #342, opened 2015-09-14
Js fixes

oparoz:
- Fix for #341 
- Implementation for #237
- Progress bar was hidden by bigshot container
- Fix for #343 

@jancborchardt , please check that the filename looks OK in the slideshow.


oparoz:
Filename in the slideshow can be improved later, so merging.

PR #348, opened 2015-09-15
adjust filename size and weight in slideshow

jancborchardt:
Make it look more in line with ownCloud. We barely use bold styles anymore, especially because it’s not needed when it’s the only text in the view.

Please review @oparoz @owncloud/designers

### DIFF
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -164,13 +164,12 @@
 #slideshow > .name .title {
 	padding: 0px 50px;
 	text-align: center;
-	color: #FFF;
-	font-size: 18px;
-	font-weight: bold;
+	color: #fff;
+	font-size: 16px;
 	line-height: normal;
 	text-overflow: ellipsis;
-	overflow:hidden;
-	white-space:nowrap;
+	overflow: hidden;
+	white-space: nowrap;
 	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
 	filter: alpha(opacity=50);
 	opacity: .5;


oparoz:
Screenshots?

I used bold because the thinness of the new font doesn't look right with the rest of the UI.
It's either we have a mismatch or a uniformed UI, but which doesn't match the style of core...

I would consider this a bug fix which can be completed past feature freeze.

oparoz:
Also, shouldn't the filename have a semi-transparent background to make it readable on a bright background?

jancborchardt:
> I used bold because the thinness of the new font doesn't look right with the rest of the UI.
> It's either we have a mismatch or a uniformed UI, but which doesn't match the style of core...

What do you mean? The style of core _is_ the thin font.

> Also, shouldn't the filename have a semi-transparent background to make it readable on a bright background?

I would rather go for a slight text-shadow (_very_ slight so it doesn’t look garish) but please let’s keep this PR to fixing the main things, ok?

oparoz:
> What do you mean? The style of core is the thin font.

The UI elements of the slideshow. The cross, the greater/smaller than signs. They're thick, not thin.
Looking at screenshots, maybe the truth lies in between...

bold
thin

oparoz:
Adding your example here. I find the filename barely readable.

jancborchardt:
That’s also because a filename has less relevance in this view than the actions. But yes, a second step would be to adjust the action icons to make them thinner to fit the general style.

But can we please stop digressing from the core here – the problem that the filename is bold and way too big?

oparoz:
> But can we please stop digressing from the core here – the problem that the filename is bold and way too big?

OK. In this PR, I find the filename too thin and unreadable, maybe keep the size 18px?

jancborchardt:
I honestly think that’s too big in relation to the image.

oparoz:
OK, we need the opinion of more designers then (as more devs opinions won't help probably).

Since it's not the first time people mention the thinness of the font, maybe it needs to be thickened slightly in a future release?

jancborchardt:
@owncloud/designers please check.

We changed the default font weight from regular to light because it gives a more modern feel. Also as discussed with @karlitschek, in case he wants to review this too.

oparoz:
Maybe change the weight in the slideshow to 500?

jancborchardt:
@oparoz that would be regular weight – we do not use that anywhere else and hence it would be highly inconsistent.

oparoz:
OK, then maybe play with the colours or other font properties to make it more legible. Dark grey and thin on a black background doesn't work so well I think.

jancborchardt:
Yeah, maybe making it a bit lighter (as in more white) will solve it.

oparoz:
I've removed the bold. I'll let you come up with the rest of the solution ;)

jancborchardt:
Great, thanks!

jancborchardt:
@oparoz after using it for a while, the size seems fine. Unbolding was the important stuff. :) Good call!

oparoz:
:+1:

PR #354, opened 2015-09-17
Retrieve the media type icon when loading an image fails

### LINKED ISSUE #346: Malformed SVG breaks gallery
A SVG file with invalid SVG breaks a whole gallery album.

```
Uncaught TypeError: Cannot set property 'alt' of null(anonymous function) @ galleryimage.js?v=58e6c41db6cb9e4292f27caf8b86a008:87(anonymous function) @ jquery.min.js?v=58e6c41db6cb9e4292f27caf8b86a008:4x.Callbacks.c @ jquery.min.js?v=58e6c41db6cb9e4292f27caf8b86a008:4x.Callbacks.p.fireWith @ jquery.min.js?v=58e6c41db6cb9e4292f27caf8b86a008:4x.extend.Deferred.x.each.i.(anonymous function) @ jquery.min.js?v=58e6c41db6cb9e4292f27caf8b86a008:4Thumbnails.loadBatch.eventSource.listen.thumb.image.onerror @ thumbnail.js?v=58e6c41db6cb9e4292f27caf8b86a008:135
```

oparoz:
Problems are usually detected on the PHP side, when trying to generate a preview, but it's not the case for SVG and there could be other problems along the road, so we fetch the media type icon in case of failure
Fixes #346

### DIFF
--- a/js/thumbnail.js
+++ b/js/thumbnail.js
@@ -132,7 +132,7 @@ function Thumbnail (fileId, square) {
 							};
 							thumb.image.onerror = function () {
 								thumb.valid = false;
-								thumb.loadingDeferred.resolve(null);
+								thumb.image.src = Thumbnails._getMimeIcon(preview.mimetype);
 							};
 
 							if (thumb.status === 200) {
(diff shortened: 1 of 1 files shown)


oparoz:
@LukasReschke - Please test with your SVGs, thanks :)

@jancborchardt - That could be the reason you got the spinning wheel in Berlin. In my tests, some SVGs wich can be loaded perfectly fine in Inkscape per example could not be loaded by the browser and would block the rendering of the photowall.

oparoz:
@setnes @demattin @mmattel Easy to test :)

jancborchardt:
For some reason it works now, and still works fine with this PR. ;)

A click on the image though leads to a 404 page:
capture du 2015-09-23 22-26-10
This should either be done via yellow notifcation, or the image not be shown at all. We shouldn’t lead people to this error page when we know the image is dead.

oparoz:
I'll look into this and will get back to you if I can't reproduce. Normally, only images with a 500 error are shown as icons, but in the case of images which were reported as valid, but failed to be downloaded, we make an exception since it's probably a rendering issue. The idea is to let the user download the file to see if it's corrupt.

oparoz:
@jancborchardt I can't reproduce your issue. A failed SVG or a broken JPG is downloaded. I need to know more to see if it's related to this PR or not.

jancborchardt:
Ok, maybe related to broken config. For me this PR doesn’t change anything so I guess it’s fine.

Please also test @deMattin @setnes! (Time to put your enthusiasm for ownCloud to work! ;)

deMattin:
OT: @jancborchardt: I will get more power again in testing after my holidays end of october. Til then, I  only can read here and annnoy you with my point of view ... ;)

oparoz:
LOL!

oparoz:
A broken SVG you can use to test:
https://oc8demo.interfacloud.com/index.php/apps/galleryplus/s/q6j106oXEF5bi2q/53739.svg

You need to enable native_svg first in your gallery.cnf

jancborchardt:
@deMattin it’s fine – maybe just try to keep your comments shorter. ;) They are valuable, but always take long to read. :D

deMattin:
Ok, I'll try it. 8)

Issue #347, opened 2015-09-15
Should use secure mimetypes instead of generic ones for download routes


LukasReschke:
See https://github.com/owncloud/core/pull/10938 for reasoning

```
GET /master/index.php/apps/gallery/files/download/3?c=b2a1325df3e7968027e36a3fe7bf8be8&requesttoken=8947b86c939d724cd60efe4a165e9129f7f7b155b41e0ea1095c84028d336537%7CyRbLmmgx%2FB7Jkpoz%7Cf80eb122cbb7f40379bdf45a3792641787bcd5d6410ce0463045a0482729ce86cc71b6f4837832488a9d7143280a93a1ca3acb0805a21b878092ed0fad4f76a0%3A4NlU4wom3c HTTP/1.1
Host: localhost
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:40.0) Gecko/20100101 Firefox/40.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Cookie: ocnqhvnh1crb=402c3ing7r53kn1dgp225i7p27; oc_sessionPassphrase=vx00ticTJMER1O8oemQbj5FK3pMGIXkUwuqNuqjjpK%2BK1jY1hiye76hB2fsoBTfarqIB3VhW6AMU0pdrIVs1wI%2FwMnygUJC%2Fwm07OGRU1gcBQU50N6%2FgNr%2FyeGfInsun
Connection: keep-alive
```

```
HTTP/1.1 200 OK
Date: Tue, 15 Sep 2015 15:34:04 GMT
Server: Apache/2.4.16 (Unix) PHP/5.6.11
X-Powered-By: PHP/5.6.11
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-cache, must-revalidate
Pragma: no-cache
Content-Security-Policy: default-src 'none';script-src 'self' 'unsafe-eval';style-src 'self' 'unsafe-inline';img-src 'self' data:;font-src 'self';connect-src 'self';media-src 'self'
Content-Disposition: attachment; filename*=UTF-8''welcome.js; filename="welcome.js"
Content-Length: 163
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
X-Robots-Tag: none
X-Frame-Options: SAMEORIGIN
[...]

PR #308, opened 2015-09-08
Protect users from gallery.cnf

oparoz:
Fixes #295 
- Enabled the sanitizing option of marked
- Added DOMPurify
- Added custom rules to DOMPurify

@LukasReschke

PR #361, opened 2015-09-18
Secure mimetypes

oparoz:
Fix for #347 
1. Always send `text/plain` for SVG at the download points
2. Only allow the SVG media type for the preview endpoint in the API if explicitly asked for. All the other preview endpoints will continue to serve `image/svg+xml` since a preview for a SVG is only asked for if NativeSVG was enabled in the config
3. Only serve files of the media types we support

@LukasReschke

### DIFF
--- a/appinfo/application.php
+++ b/appinfo/application.php
@@ -90,7 +90,6 @@ class Application extends App {
 				$c->query('AppName'),
 				$c->query('Request'),
 				$c->query('ConfigService'),
-				$c->query('PreviewService'),
 				$c->query('Logger')
 			);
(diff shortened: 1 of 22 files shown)


oparoz:
Ping @LukasReschke

LukasReschke:
Pong. Let me look at this today :smile:

oparoz:
Thanks! I've rebased the other one, so you can have a look as well :) https://github.com/owncloud/gallery/pull/308

LukasReschke:
:+1: - THX!

PR #367, opened 2015-09-23
Import legacy share dialog code

PVince81:
This is to make the share dropdown work again
- [x] TODO: fix checkboxes

Let's wait for the checkboxes to be fixed in https://github.com/owncloud/calendar/pull/940 then grab the share.js file from there again.

@oparoz @blizzz


karlitschek:
thanks for the fix aka workaround :+1:

PVince81:
I have updated the code, but there seem to be additional issues with the checkboxes @Henni 
checkbox-issue-gallery-share

This wasn't visible in the calendar app because there was no link share.
Best would be to search for all occurences of "checkbox" in that file and make the change.
Hint: there are other hidden checkboxes that only appear when setting options in the "Admin" page

oparoz:
Thank you very much for this @PVince81 :)

PVince81:
@Henni do you intend to fix the checkboxes here too ?
Another alternative is reverting them back to the old style, basically grabbing share.js before your changes.

Henni:
I'll take a look at this.
Does this issue still exist or are the old checkboxes properly displayed?

oparoz:
> Does this issue still exist or are the old checkboxes properly displayed?

Has there been a change merged in master recently to fix this?

Henni:
@oparoz https://github.com/owncloud/core/pull/19329

oparoz:
Thanks @Henni 

So, the checkboxes look like the old ones and I'm guessing we need to fix the CSS in another PR by adding the new class to the elements and possibly modify them?

Apart from that, it all seems to work, apart from mail notifications, but which never worked. Good job @PVince81.

PVince81:
Ok then I guess I'll grab the old share.js version which didn't have half of the checkboxes moved to the new style. So in the end the legacy share dropdown will only have old checkboxes.

PVince81:
@oparoz I have replaced the code in this PR with the older share.js, the one without half the checkboxes changed. I suggest to merge this first.

If you want the new checkbox styles this can be done in a separate PR.

oparoz:
Is it OK if I move that code to the vendor folder? It's like a library to me.

PVince81:
As you like, feel free to push here

oparoz:
OK. Any idea why "sending an email" never shows up when using that dialogue? Is it working in the other apps?

oparoz:
@henni, seems like there is a bug with the fallback

fallback_not_working

oparoz:
Rebased

icewind1991:
Fixed password protect checkbox



With password protect enforced:

oparoz:
Report from @Dianafg76

| Test case | Browser | Result |
| --- | --- | --- |
| Share from the Gallery a folder | Firefox, Chrome, Safary | OK |
| Share with password | Firefox, Chrome, Safary | OK |
| Share with expiration time | Firefox, Chrome, Safary | OK |
| Share with password and expiration time | Firefox, Chrome, Safary | OK |
| Force from admin to share with password | Firefox, Chrome | OK |
| Force from admin to share with password | Safary | NOK (https://github.com/owncloud/gallery/issues/399) |

PR #285, opened 2015-08-31
Bouncy scroll for iOS devices

ghost:
Scroll more smothly for webkit browser(mobile)

PR #368, opened 2015-09-23
improve iOS scrolling as per @fuhaoshih's pull request #285

jancborchardt:
PR superseding https://github.com/owncloud/gallery/pull/285

(Couldn’t cherry-pick because of the conflicts.) Please review @oparoz @owncloud/designers @fuhaoshih

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -75,6 +75,7 @@ div.crumb.last a {
 	-o-transition: background-color 1s linear;
 	-ms-transition: background-color 1s linear;
 	transition: background-color 1s linear;
+	-webkit-overflow-scrolling: touch;
 }
 
 #gallery.hascontrols {


jancborchardt:
@mmattel please review too.

oparoz:
It's been approved in the other PR already, but I'm curious. Has anybody tested this on other browsers to see if it has any side effects? It shouldn't, but I'd rather be sure.

mmattel:
I have checked this, but it seems that the css parameter does not get triggered, maybe positioned on the wrong place, but I am not a css expert. 

I have tested on my iPad following page where you can test this with and without and there it works:
https://css-tricks.com/snippets/css/momentum-scrolling-on-ios-overflow-elements/

I have used private browsing mode and also cleared the iPad Safari cache ect. but without success.

It also seems according to google search that an overflow parameter hast to be set too, but this also did not change anything in my test...

```
overflow-y: scroll;                 /* has to be scroll, not auto */
-webkit-overflow-scrolling: touch;
```

I would appreciate if that would work :smile:

oparoz:
It breaks non-Safari browsers, so :-1: 
https://github.com/owncloud/core/issues/19334

I'm sure there are ways to make it work without breaking stuff, but apparently, it's going to take more effort than that one line.

jancborchardt:
Closing then, thanks for the catch @oparoz

PR #369, opened 2015-09-23
improve touch area of slideshow controls, fix #53

### LINKED ISSUE #53: Improve the touch area of some buttons
I think _next_ and _previous_ should have larger click/tap areas.
Buttons at the top (download, toggle, etc.) should also have a larger tap area and should be spaced accordingly. 
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9060550-improve-the-touch-area-of-some-buttons?utm_campaign=plugin&utm_content=tracker%2F9328526&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F9328526&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jancborchardt:
Fix https://github.com/owncloud/gallery/issues/53, please review @oparoz @owncloud/designers 

Improvements:
- »Next« button takes all space, so clicking on the image now also advances it (as with most Galleries). Because it’s first in the DOM, it will be overlayed by the other actions where they are.
- »Previous« button takes up 20% of the left side
- »Exit« and »Play/Pause« are both bigger

### DIFF
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -93,16 +93,19 @@
 }
 
 #slideshow > .next, #slideshow > .previous {
-	top: 50%;
-	margin-top: -21px;
+	height: 100%;
+	top: 0;
+	margin-top: 0;
 }
-
 #slideshow > .next {
+	width: 80%;
 	right: 0;
+	background-position: right 10px center;
 }
-
 #slideshow > .previous {
+	width: 20%;
 	left: 0;
+	background-position: 10px center;
 }
 
 #slideshow > .exit {
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -90,6 +90,7 @@
 	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=75)";
 	filter: alpha(opacity=75);
 	opacity: .75;
+	transition: opacity 300ms;
 }
 
 #slideshow > .next, #slideshow > .previous {
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -46,8 +46,10 @@
 }
 
 #slideshow.inactive > input,
+#slideshow.inactive > input:hover,
 #slideshow.inactive > .progress,
 #slideshow.inactive > .menu input,
+#slideshow.inactive > .menu input:hover,
 #slideshow.inactive > .name {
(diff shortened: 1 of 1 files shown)


karlitschek:
makes sense :+1:

oparoz:
> »Exit« and »Play/Pause« are both bigger

I'm a bit worried about this, because when other buttons are going to be added there won't be much room for them if they each need to be 100px wide.

jancborchardt:
@oparoz I was also not sure about those two because they already have a lot of space. Will revert those back to original state, is this ok then?

oparoz:
> is this ok then?

In theory, yes, now I need to thoroughly test it to make sure there aren't any interferences, because it has happened in the past :)

jancborchardt:
@oparoz fixed :) Please check again, cc @owncloud/designers

oparoz:
Unfortunately, there is a strange side-effect, at least on Firefox.
1. Launch the slideshow
2. Let the controls fade away
3. Bring the cursor over the slideshow

All the controls come back together, except for next...

jancborchardt:
@Henni @MorrisJobke @owncloud/designers can you help fix this so we can get this in? I’m on vacation.

oparoz:
I'm baffled by this, the fact that it's fatter makes it slower...

Henni:
The problem was that the hover state of all controls sets a opacity.
This opacity transitions based on the inactive class with a speed of `3s`.
After the hover transition is triggered the `.inactive` class is removed and all other controls trigger to the normal opacity within `300ms`.

I fixed this by adding a transition speed to the hover rule.

MorrisJobke:
Tested and looks good :+1:

oparoz:
Thanks @henni.

There is another side-effect, which is highlighted by this change.
1. Launch the slideshow
2. Leave the mouse pointer anywhere on the screen (center to right area)
3. Wait 3s

The `next` button just stays there. It happens to all the buttons if you leave the mouse pointer over them and it's not that bad with other buttons, since you're really close to the action, but with this change, the next button stays there in most cases, forcing the user to move the pointer out of the window in order to get rid of it.

Thoughts?

Henni:
@oparoz what do you think of this fix?

oparoz:
All good. Thank you very much for your help @Henni !

PR #379, opened 2015-09-24
[8.2, 8.1] Remove obsolete preview validator

oparoz:
Things have improved in core and we now get previews of the correct size.
### Tests
1. Clear your thumbnails
2. Always load a folder which has sub-albums and pay attention to the small thumbnails in the sub-albums
3. Entering the sub-albums to make sure the regular thumbnails also look OK.
#### Plan

Test with:
- square pictures
- wide pictures
- tall pictures


oparoz:
@jancborchardt @karlitschek - I'm guessing this is the sort of things that should wait for 9.0?

karlitschek:
Your call :-) But maybe something for 9.0

oparoz:
OK. I think it would be best to get rid of it before making the app official. Less code to maintain/audit.

Updated the OP, with a few things to check and I'll decide later, if enough people have tested the removal of the feature.

PR #384, opened 2015-09-25
increase filename visibility, same as for icons in slideshow

jancborchardt:
Similar to the icon visibility fix in core https://github.com/owncloud/core/pull/19380, this also makes the filename visible on bright background.

Please review @owncloud/designers @oparoz @deMattin @setnes

### DIFF
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -162,17 +162,17 @@
 }
 
 #slideshow > .name .title {
-	padding: 0 50px;
+	padding: 0 50px 3px;
 	text-align: center;
-	color: #FFF;
+	color: #fff;
+	text-shadow: 0 0 10px #333;
 	font-size: 18px;
 	line-height: normal;
 	text-overflow: ellipsis;
 	overflow: hidden;
 	white-space: nowrap;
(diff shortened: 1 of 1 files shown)


karlitschek:
:+1:

oparoz:
I'd say: almost

If I compare the filename with the icons, it seems the border is missing, which makes it difficult to read on white.

no_border

jancborchardt:
@oparoz there’s no such thing as text-border unfortunately. I’d say it’s better than nothing and we should merge. (Hello from vacation btw, this is an exception cc @owncloud/designers you need to handle this the next 2 weeks. ;)

stefan-niedermann:
> there’s no such thing as text-border unfortunately

Nope, but you can use text-shadow to simulate a text-border.

``` css
    text-shadow:  1px  1px 1px black,
         1px -1px 1px black,
        -1px  1px 1px black,
        -1px -1px 1px black;
```

oparoz:
Thanks for the tip @stefan-niedermann .

@deMattin @setnes @karlitschek  Please test the updated PR. I think the title now looks similar to the icons :)

@jancborchardt Enjoy the time off :)

icewind1991:
Looks good :+1:

PR #387, opened 2015-09-27
New occ commands to manage thumbnails

oparoz:
Sunday Funday!

Here are some command some of you may find useful to manage your thumbnails: 
## Create thumbnails

Solution for https://github.com/owncloud/core/issues/17536
### Documentation

[Create thumbnails](https://github.com/owncloud/gallery/wiki/Create-Thumbnails)
### Specs
- Collect user(s) or path info from the list of command arguments
- Die if encryption is enabled
- Compute path to analyse
- Only work with folders starting with `/<user>/files`
- Scan the given path using Cache\Scanner in order to work with reliable data
- Update the DB cache
- Generate previews **only for media types supported by the Gallery and for which we have a preview provider**
- Collect stats
#### Stats
- # of folders
- # of files
- # of images
- # of previews generated
- # of failed previews (listed)
- Time to complete request
- Last scanned file
- Size of processed data
### TODO
- [x] Quit if encryption is enabled
- [x] Only scan a mounted volume if previews are enabled for that storage location
- [x] Restrict scan to `/<user>/files`
- [x] Add some stats
- [x] Add a switch to force the regeneration of thumbnails
- [x] ~~Offer a safer scanning option, with transactional locking. I'm hoping this will solve the error 1205 problems and not create more problems when trying to generate thumbnails~~
- [x] Presents stats even if operation was cancelled
- [ ] Add tests 
- [ ] Make operation cancelling optional
- [ ] Use [OCP\Files\Cache\IPropagator](https://github.com/owncloud/core/blob/master/lib/public/files/cache/ipropagator.php) on 9
- [ ] Check if this is fixed https://github.com/owncloud/core/issues/17196
### Possible extensions

For future PRs
- Make coffee
- Support mimetype as a filter
### Documentation

[Delete thumbnails](https://github.com/owncloud/gallery/wiki/Delete-Thumbnails)
### Specs
- Collect user(s) or path info from the list of command arguments
- We do not care about encryption
- Wipe the thumbnails folders if we're asked to remove thumbnails for one, a few or all users
- Compute path to analyse in all other cases
- Search for files **of media type supported by Gallery and for which we have a preview provider** and delete as we go
- Collect stats
#### Stats
- # of images
- # of thumbnails deleted
- # of failed deletion (listed)
- Time to complete request
- Saved space
### Possible extensions

For future PRs
- There could be an option to delete thumbnails for all files. This would only be useful if the providers were changed in between the time the thumbnails were generated and the time at which the command is run
### TODO
- [x] Find an easy way to update the cache without re-scanning
- [x] Fix bug in core which prevents the use of a shared folder as the path (sub-folders work) https://github.com/owncloud/core/issues/19544 (8.1+)
- [x] Ask confirmation before launching the operation as we're deleting files from the system
- [ ] Add tests 
## Notes

Of course, it would be much more useful to be able to run that from a cron, from the user's settings panel, etc. so don't hesitate to improve this solution ;)

@mmattel @setnes @EricReiche @jospoortvliet @DeepDiver1975 @icewind1991 @Xenopathic @demattin @thedd @patman15 @africanforest @MorrisJobke @PVince81 @libasys @mbazdell @blackm

### DIFF
--- a/command/thumbnails.php
+++ b/command/thumbnails.php
@@ -1,0 +1,473 @@
+<?php
+/**
+ * ownCloud - gallery
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Bart Visscher <bartv@thisnet.nl>
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
+ * @author Morris Jobke <hey@morrisjobke.de>
+ * @author Robin Appelman <icewind@owncloud.com>
+ * @author Vincent Petry <pvince81@owncloud.com>
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Olivier Paroz <owncloud@interfasys.ch>
+ *
+ * @copyright Copyright (c) 2015, ownCloud, Inc.
+ * @copyright Olivier Paroz 2015
+ */
+
+namespace OCA\Gallery\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+// Static methods needed are not available in OCP
+use OC\Files\Filesystem;
+
+use OCP\IUserManager;
+use OCP\IUser;
+use OCP\Files\IRootFolder;
+use OCP\Files\Mount\IMountManager;
+use OCP\Files\Mount\IMountPoint;
+use OCP\Files\Node;
+
+use OCA\Gallery\Service\ConfigService;
+
+/**
+ * Class Thumbnails
+ *
+ * Manages thumbnails of visual media types supported by the system
+ *
+ * @package OCA\Gallery\Command
+ */
+abstract class Thumbnails extends Command {
+
+	/** @var IUserManager */
+	protected $userManager;
+	/** @var IRootFolder */
+	protected $rootFolder;
+	/** @var IMountManager */
+	protected $mountManager;
+	/** @var ConfigService */
+	protected $configService;
+	/** @var array */
+	protected $supportedMediaTypes;
+	/** @var float */
+	protected $execTime = 0;
+	/** @var int */
+	protected $folderSize = 0;
+	/** @var int */
+	protected $foldersCounter = 0;
+	/** @var int */
+	protected $filesCounter = 0;
+	/** @var int */
+	protected $imagesCounter = 0;
+	/** @var string */
+	protected $lastFile = '';
+	/** @var string */
+	protected $lastFolder = '';
+	/** @var int */
+	protected $operationCounter = 0;
+	/** @var int */
+	protected $failedOperationCounter = 0;
+	/** @var array */
+	protected $failedOperations;
+	/** @var bool */
+	protected $interrupted = false;
+
+	/**
+	 * Constructor
+	 *
+	 * @param IUserManager $userManager
+	 * @param IRootFolder $rootFolder
+	 * @param IMountManager $mountManager
+	 * @param ConfigService $configService
+	 */
+	public function __construct(
+		IUserManager $userManager,
+		IRootFolder $rootFolder,
+		IMountManager $mountManager,
+		ConfigService $configService
+	) {
+		$this->userManager = $userManager;
+		$this->rootFolder = $rootFolder;
+		// IMountManager is 8.2+ only
+		$this->mountManager = $mountManager;
+		$this->configService = $configService;
+		// We want all the media types supported via providers only
+		$this->supportedMediaTypes = $this->configService->getSupportedMediaTypes(true, false);
+
+		parent::__construct();
+	}
+
+	/**
+	 * Processes PHP errors as exceptions in order to be able to keep track of problems
+	 *
+	 * @see https://secure.php.net/manual/en/function.set-error-handler.php
+	 *
+	 * @param int $severity the level of the error raised
+	 * @param string $message
+	 * @param string $file the filename that the error was raised in
+	 * @param int $line the line number the error was raised
+	 *
+	 * @throws \ErrorException
+	 */
+	public function exceptionErrorHandler($severity, $message, $file, $line) {
+		if (!(error_reporting() & $severity)) {
+			// This error code is not included in error_reporting
+			return;
+		}
+		throw new \ErrorException($message, 0, $severity, $file, $line);
+	}
+
+	/**
+	 * Initialises some useful tools for the Command
+	 */
+	protected function initTools() {
+		// Start the timer
+		$this->execTime = -microtime(true);
+
+		// Convert PHP errors to exceptions
+		set_error_handler([$this, 'exceptionErrorHandler'], E_ALL);
+
+		// Collect interrupts and notify the running command
+		pcntl_signal(SIGTERM, [$this, 'cancelOperation']);
+		pcntl_signal(SIGINT, [$this, 'cancelOperation']);
+	}
+
+	/**
+	 * Parses all the parameters submitted by the command user
+	 *
+	 * @param InputInterface $input
+	 * @param string $userInputErrorMsg
+	 * @param string $pathInputErrorMsg
+	 * @param OutputInterface $output
+	 *
+	 * @return array
+	 * @throws InputCommandException
+	 */
+	protected function parseParameters(
+		InputInterface $input, $userInputErrorMsg, $pathInputErrorMsg, OutputInterface $output
+	) {
+		$inputPath = $input->getOption('path');
+		if ($inputPath) {
+			$inputPath = '/' . trim($inputPath, '/');
+			try {
+				list (, $user, $topFolder) = explode('/', $inputPath, 4);
+				$users = [$user];
+			} catch (\ErrorException $exception) {
+				throw new InputCommandException($pathInputErrorMsg);
+			}
+			$this->checkPathParameter($topFolder, $pathInputErrorMsg);
+		} else if ($input->getOption('all')) {
+			$users = $this->userManager->search('');
+		} else {
+			$users = $input->getArgument('user_id');
+		}
+		$this->checkUserParameter($users, $userInputErrorMsg);
+
+		return [$inputPath, $users];
+	}
+
+	/**
+	 * Presents the results of the operation in tabular form
+	 *
+	 * @param OutputInterface $output
+	 */
+	protected function presentResults(OutputInterface $output) {
+		// Stop the timer
+		$this->execTime += microtime(true);
+
+		$output->setVerbosity(OutputInterface::VERBOSITY_NORMAL);
+		$output->writeln("");
+	}
+
+	/**
+	 * Formats microtime into a human readable format
+	 *
+	 * @return string
+	 */
+	protected function formatExecTime() {
+		list($secs, $tens) = explode('.', sprintf("%.1f", ($this->execTime)));
+		$niceDate = date('H:i:s', $secs) . '.' . $tens;
+
+		return $niceDate;
+	}
+
+	/**
+	 * Formats bytes into a human readable format
+	 *
+	 * @param int $bytes
+	 * @param int $decimals
+	 *
+	 * @return string
+	 * @see https://secure.php.net/manual/en/function.filesize.php#106569
+	 */
+	protected function formatSize($bytes, $decimals = 2) {
+		$size = 'BKMGTP';
+		$factor = floor((strlen($bytes) - 1) / 3);
+
+		return sprintf("%.{$decimals}f", $bytes / pow(1024, $factor)) . @$size[$factor];
+	}
+
+	/**
+	 * Shows a summary of operations
+	 *
+	 * @param string[] $headers
+	 * @param string[] $rows
+	 * @param OutputInterface $output
+	 */
+	protected function showSummary($headers, $rows, OutputInterface $output) {
+		$niceDate = $this->formatExecTime();
+		$size = $this->formatSize($this->folderSize);
+		if (!$rows) {
+			$rows = [
+				$this->foldersCounter,
+				$this->filesCounter,
+				$this->imagesCounter,
+				$this->operationCounter,
+				$this->failedOperationCounter,
+				$niceDate,
+				$size
+			];
+		}
+		$table = new Table($output);
+		$table
+			->setHeaders($headers)
+			->setRows([$rows]);
+		$table->render();
+
+		$output->writeln("Last scanned file <info>$this->lastFile</info>");
+	}
+
+	/**
+	 * Shows details about failed operations
+	 *
+	 * @param string[] $headers
+	 * @param OutputInterface $output
+	 */
+	protected function showFailed($headers, OutputInterface $output) {
+		if (count($this->failedOperations)) {
+			$table = new Table($output);
+			$table->setHeaders($headers);
+			foreach ($this->failedOperations as $failedOperation) {
+				$table->addRow([$failedOperation]);
+			}
+			$table->render();
+		}
+	}
+
+	/**
+	 * Performs a given operation on a given user's file system
+	 *
+	 * @param string $scanOperation
+	 * @param string $user
+	 * @param string $dir
+	 * @param OutputInterface $output
+	 *
+	 * @return array
+	 * @throws ForbiddenCommandException|InputCommandException
+	 * @see OC\Files\Utils\Scanner::scan
+	 */
+	protected function performOperation($scanOperation, $user, $dir, OutputInterface $output
+	) {
+		if (!Filesystem::isValidPath($dir)) {
+			throw new InputCommandException('Invalid path to scan');
+		}
+		$mounts = $this->getMounts($dir, $user);
+		$result = [];
+		foreach ($mounts as $mount) {
+			$previewsAllowed = $mount->getOption('previews', true);
+			if (is_null($mount->getStorage()) || !$previewsAllowed) {
+				continue;
+			}
+			$storage = $mount->getStorage();
+			// if the home storage isn't writable then the scanner is run as the wrong user
+			if ($storage->instanceOfStorage('\OCP\Files\IHomeStorage') and
+				(!$storage->isCreatable('') or !$storage->isCreatable('files'))
+			) {
+				throw new ForbiddenCommandException(
+					"<error>Home storage for user $user not writable</error>\nMake sure you're running the scan command only as the user the web server runs as"
+				);
+			}
+			// We can now execute the wanted operation
+			$result[] = call_user_func_array(
+				[$this, $scanOperation],
+				[$storage, $mount, $dir, $user, $output]
+			);
+		}
+
+		return $result;
+	}
+
+	/**
+	 * Determines if we have a file
+	 *
+	 * @param Node $node
+	 *
+	 * @return bool
+	 */
+	protected function isFile(Node $node) {
+		$isFile = false;
+		// TODO: Find a more reliable way as we can't trust FileInfo
+		if ($node->getType() === 'file') {
+			$isFile = true;
+		}
+
+		return $isFile;
+	}
+
+	/**
+	 * Makes sure the folder or file exists in the user's filesystem
+	 *
+	 * @param string $path a full path or a relative path starting from the user's folder
+	 * @param string|null $user
+	 *
+	 * @return bool|Node
+	 *
+	 */
+	protected function getNode($path, $user = null) {
+		if ($user) {
+			$path = '/' . $user . '/' . $path;
+		}
+		if ($this->rootFolder->nodeExists($path)) {
+			return $this->rootFolder->get($path);
+		}
+
+		return false;
+	}
+
+	/**
+	 * @return bool
+	 */
+	protected function hasBeenInterrupted() {
+		$cancelled = false;
+		pcntl_signal_dispatch();
+		if ($this->interrupted) {
+			$cancelled = true;
+		}
+
+		return $cancelled;
+	}
+
+	/**
+	 * Asks for confirmation before proceeding
+	 *
+	 * @param InputInterface $input
+	 * @param string $question
+	 * @param OutputInterface $output
+	 *
+	 * @throws InputCommandException
+	 */
+	protected function getConfirmation(InputInterface $input, $question, OutputInterface $output) {
+		$helper = $this->getHelper('question');
+		$confirmationQuestion = new ConfirmationQuestion($question, false);
+
+		$quiet = false;
+		if ($output->getVerbosity() === 0) {
+			$quiet = true;
+			$output->setVerbosity(OutputInterface::VERBOSITY_NORMAL);
+		}
+		if (!$helper->ask($input, $output, $confirmationQuestion)) {
+			throw new InputCommandException("Operation aborted");
+		}
+		if ($quiet) {
+			$output->setVerbosity(OutputInterface::VERBOSITY_QUIET);
+		}
+	}
+
+	/**
+	 * Tracks failed operations
+	 *
+	 * @param string $path
+	 * @param string $message
+	 * @param OutputInterface $output
+	 */
+	protected function trackFailedOperation($path, $message, OutputInterface $output) {
+		$this->failedOperationCounter++;
+		$this->failedOperations[] = $path;
+		$this->printToConsole("<error>$message</error>", $output);
+	}
+
+	/**
+	 * Prints a message to the console if we're not in quiet mode
+	 *
+	 * @param string $message
+	 * @param OutputInterface $output
+	 */
+	protected function printToConsole($message, OutputInterface $output) {
+		if (!$output->getVerbosity() === 0) {
+			$output->writeln($message);
+		}
+	}
+
+	/**
+	 * Changes the status of the command to "interrupted"
+	 *
+	 * Gives a chance to the command to properly terminate what it's doing
+	 */
+	private function cancelOperation() {
+		$this->interrupted = true;
+	}
+
+	/**
+	 * Makes sure the user has entered a valid path
+	 *
+	 * @param IUser[]|string[] $users
+	 * @param string $userInputErrorMsg
+	 *
+	 * @throws InputCommandException
+	 */
+	private function checkUserParameter($users, $userInputErrorMsg) {
+		if (count($users) === 0) {
+			throw new InputCommandException($userInputErrorMsg);
+		}
+	}
+
+	/**
+	 * Makes sure the user has entered a valid path
+	 *
+	 * It's too early to make sure the path is valid as we'll only mount the filesystem later in
+	 * the process
+	 *
+	 * @param string $pathInputErrorMsg
+	 *
+	 * @throws InputCommandException
+	 */
+	private function checkPathParameter($topFolder, $pathInputErrorMsg) {
+		if ($topFolder !== 'files') {
+			throw new InputCommandException($pathInputErrorMsg);
+		}
+	}
+
+	/**
+	 * Retrieves all storage mount points in $dir, including external ones
+	 *
+	 * @param string $dir
+	 * @param string $user
+	 *
+	 * @return IMountPoint[]
+	 * @throws InputCommandException
+	 * @see OC\Files\Utils\Scanner::getMounts
+	 */
+	private function getMounts($dir, $user) {
+		\OC_Util::tearDownFS();
+		\OC_Util::setupFS($user);
+
+		$testPath = $this->rootFolder->nodeExists($dir);
+		if ($testPath) {
+			$mounts = $this->mountManager->findIn($dir);
+			$mounts[] = $this->mountManager->find($dir);
+			$mounts = array_reverse($mounts); //start with the mount of $dir
+
+			return $mounts;
+		} else {
+			throw new InputCommandException("The path provided does not exist");
+		}
+	}
+
+}
--- a/appinfo/register_command.php
+++ b/appinfo/register_command.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * ownCloud - gallery
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Olivier Paroz <owncloud@interfasys.ch>
+ * @author Bernhard Posselt <dev@bernhard-posselt.com>
+ *
+ * @copyright Olivier Paroz 2015
+ * @copyright Bernhard Posselt 2015
+ */
+
+use OCP\AppFramework\IAppContainer;
+
+if (\OC::$server->getConfig()
+				->getSystemValue('installed', false)
+) {
+	$galleryApp = new OCA\Gallery\AppInfo\Application();
+	$galleryContainer = $galleryApp->getContainer();
+	$galleryContainer->registerService(
+		'userFolder', function (IAppContainer $c) {
+		return null;
+	}
+	);
+	$galleryContainer->registerService(
+		'OCP\Encryption\IManager', function (IAppContainer $c) {
+		return $c->getServer()
+				 ->getEncryptionManager();
+	}
+	);
+	$galleryContainer->registerService(
+		'OCP\Files\Mount\IMountManager', function (IAppContainer $c) {
+		return $c->getServer()
+				 ->getMountManager();
+	}
+	);
+	$galleryContainer->registerService(
+		'OCP\Lock\ILockingProvider', function (IAppContainer $c) {
+		return $c->getServer()
+				 ->getLockingProvider();
+	}
+	);
+
+	$createThumbnailsCmd = $galleryContainer->query('OCA\Gallery\Command\CreateThumbnails');
+	$deleteThumbnailsCmd = $galleryContainer->query('OCA\Gallery\Command\DeleteThumbnails');
+
+	$application->add($createThumbnailsCmd);
+	$application->add($deleteThumbnailsCmd);
+}
--- a/command/commandexception.php
+++ b/command/commandexception.php
@@ -1,0 +1,18 @@
+<?php
+/**
+ * ownCloud - gallery
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Olivier Paroz <owncloud@interfasys.ch>
+ *
+ * @copyright Olivier Paroz 2015
+ */
+
+namespace OCA\Gallery\Command;
+
+/**
+ * Thrown when the command is not authorised to access a file
+ */
+class CommandException extends \Exception {}
--- a/command/createthumbnails.php
+++ b/command/createthumbnails.php
@@ -1,0 +1,461 @@
+<?php
+/**
+ * ownCloud - gallery
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Bart Visscher <bartv@thisnet.nl>
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
+ * @author Morris Jobke <hey@morrisjobke.de>
+ * @author Robin Appelman <icewind@owncloud.com>
+ * @author Vincent Petry <pvince81@owncloud.com>
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Olivier Paroz <owncloud@interfasys.ch>
+ *
+ * @copyright Copyright (c) 2015, ownCloud, Inc.
+ * @copyright Olivier Paroz 2015
+ */
+
+namespace OCA\Gallery\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+// Does not exist in OCP
+use OC\Files\Cache\ChangePropagator;
+use OC\Files\Storage\Storage;
+// Does not exist in OCP and needed by ChangePropagator
+use OC\Files\View;
+// Methods we need are not available in OCP
+use OC\Preview;
+
+use OCP\IDBConnection;
+use OCP\IUserManager;
+use OCP\IUser;
+use OCP\Files\IRootFolder;
+// 8.2+ only
+use OCP\Files\Mount\IMountManager;
+use OCP\Files\Mount\IMountPoint;
+use OCP\Files\File;
+use OCP\Files\Folder;
+use OCP\Encryption\IManager;
+use OCP\Lock\ILockingProvider;
+
+use OCA\Gallery\Service\ConfigService;
+
+/**
+ * Class CreateThumbnails
+ *
+ * Creates missing thumbnails of visual media types supported by the system
+ *
+ * @package OCA\Gallery\Command
+ */
+class CreateThumbnails extends Thumbnails {
+
+	/** @var GalleryScanner */
+	private $scanner;
+	/** @var IManager */
+	private $encryptionManager;
+	/** @var ChangePropagator */
+	private $propagator;
+	/** @var int */
+	private $thumbnailWidth = 400;
+	/** @var int */
+	private $thumbnailHeight = 200;
+	/** @var bool */
+	private $regenerate = false;
+
+	/**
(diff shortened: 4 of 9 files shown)


deMattin:
:+1:

mmattel:
:+1: 
One additional parameter would be great `--mount=mount_name` (with the ability in combination with --all, or general for all). Then you can add a mount with files external and (after scanning) you can run the thumbnail creation immediately for exactly and only this mount for (all) users...

oparoz:
@mmattel - If you know the path to the mount, you can already use `--path`. Let me know if it doesn't work. It would also be included in the `--all` command

oparoz:
I've just tested with a mounted ownCloud instance and it worked. I'll updated the OP.

icewind1991:
From what I can tell it only creates thumbnails for a single size which wont be enough for people with HiDPI screens, maybe add some sort of scale parameter to handle that?

oparoz:
> From what I can tell it only creates thumbnails for a single size which wont be enough for people with HiDPI screens, maybe add some sort of scale parameter to handle that?

The main goal is to generate the max preview, but since I need to also give another size (there is no getMaxPreview() method), I picked the Gallery thumbnail dimensions with a pixel ratio of 1 since we can't guess what the ratio is going to be in the various clients connecting to the app.
I could make some modifications in Gallery to always round up the ratio and offer the option to generate photowall thumbnails at a ratio of 1,2,3 and 4.

If `core` ever gets standardised sizes for thumbnails (36,256,1024,2048,4096 per example), there could be a switch to generate all of those, but currently it wouldn't make sense. If we take into consideration the various pixel ratios, we would end up having to pre-generate 20 thumbnails per image.

mmattel:
@oparoz 

Testing if creating thumbnails of a mount with --path works, here is the result:

Having a clean fresh install from git master and finalizing it.
Adding the PR to test it
Checking if ./occ has the newly added parameter -- ok
Enabling External Storage Support
Adding a SMB share named "SMB"

`sudo -u www-data ./occ gallery:create-thumbnails --path=SMB`
--> Unknown user SMB

This is what I expected, because having a mount only is not a path having a user in. And the mounts are not in the users home...

But this is what my suggestion is about. Usually lots of picture files are mounted with external storage support... And advanced thumbnail creation on an mount would be a big help.
`sudo -u www-data ./occ gallery:create-thumbnails --mount=SMB`

icewind1991:
`./occ gallery:create-thumbnails --path=/myuser/files/SMB` ?

mmattel:
@icewind1991 @oparoz 
YES :smile: this works. Thanks for the tip!
Pls document this ability because it is not explicit clear from the example above.

Question: do I get it right, that --all and --path do exclude each other?
Having this would be perfect btw... (if the first path parameter (user) is not found and --all is not present than error else do for all)
Ok, I know, many things, but this is really a great command !!

mmattel:
@oparoz 
For your info, when starting the command, I get this printed on the terminal before the generation starts. After 2-3 seconds, generation ist starting fine.

```
tdb(__NULL__): tdb_open_ex: called with name == NULL
tdb(__NULL__): tdb_open_ex: called with name == NULL
tdb(__NULL__): tdb_open_ex: called with name == NULL
tdb(__NULL__): tdb_open_ex: called with name == NULL
```

mmattel:
@icewind1991 
I think that the following is not related to this PR, but don´t you work on an issue like that?
Maybe this helps finding/fixing the problem...

After a short time running the command, I get on each thumb created a message like below logged, and I have thousands of pictures to be processed:

```
{"reqId":"G+T\/voqBTE73dsav+oi5","remoteAddr":"","app":"no app in context","mes
sage":"Trying to acquire a lock for 'files\/0d1cc4de07964c1a4c92887962ccef01' w
hile inside a transition","level":2,"time":"2015-09-28T17:13:27+00:00"}
```

oparoz:
> Question: do I get it right, that --all and --path do exclude each other?

Yes, the user will be determined from the path

> tdb(**NULL**): tdb_open_ex: called with name == NULL

Samba error. Perhaps it's scanning a folder it doesn't have access to, like some cache? You can debug this with the files:scan command on which this is based.

oparoz:
@mmattel - I investigated a bit and pushed some changes.
- Folders where being sent as files, so that's never good and I've included an extra check
- The `scanFile` event fires before a file is actually ready to be processed and `scanFilePost` almost never fired for me. I've submitted a PR to fix that (https://github.com/owncloud/core/pull/19460). If that's merged, then I'll be able to improve the scanner

mmattel:
@oparoz 
Q: will your ongoing changes also fix that temp files ceated are immediatley released after processing the file, pls see my comments about that  https://github.com/owncloud/core/issues/19444#issuecomment-144119968

oparoz:
@mmattel I'm not sure anything clean can be done about this. GD seems to be using the temp folder to process images and I don't have a command or the filenames.
A hack could get rid of `*.JPG` files after each operation.

oparoz:
OK, so I've imported most of the Scanner class in order to customise the scan since the original class didn't give us access to the mount options before scanning that storage location.

Using the latest master, we can now listen to scanPostFile event (https://github.com/owncloud/core/pull/19460) and the Image class cleans up after itself (https://github.com/owncloud/core/pull/19471), so the command should behave as expected.

@mmattel Let us know how it goes with your 19K images :)

mmattel:
Just for the record, this is the fix to my comment above regarding temp space usage: https://github.com/owncloud/core/pull/19471 (Clean temp files used for thumbnail generation, merged)

oparoz:
I'm running it on an external SMB sharer and I think we needs some stats...
- # of folders
- # of files
- # of previews generated
- # of failed previews, with list probably
- Time to complete request
- Last scanned file

oparoz:
stats

oparoz:
Summary stats
stats

oparoz:
smb_stats

I will take an hour with a large collection...

mmattel:
Info: I let it run over my >18700 pictures and it suceeded over night :smile: 

When I tested with the browser to access them via gallery, it again took time to show them, like they are created the first time... Luckily I checked the number of files and directories in advance in the thumbnail directory:

```
Before browser/gallery access, directly after the finish of ./occ
xxx/thumbnails$ tree . | tail -1
  14174 directories, 28584 files

After browser/gallery access (some picture folders accessed)
xxx/thumbnails$ tree . | tail -1
  14174 directories, 29191 files
```

This means for me that my experience was correct, the ./occ created thumbs were not used but additional once created on access. Why havn´t the ./occ created thumbs been used?

Could it be that the sizes created are different? Maybe the browser access requests the creation of a particular size and the ./occ command creates another size...

oparoz:
> When I tested with the browser to access them via gallery, it again took time to show them

It's SMB and the latency is higher than with local folders. Gallery first makes sure the file is available before asking for a thumbnail.

> Could it be that the sizes created are different? Maybe the browser access requests the creation of a particular size and the ./occ command creates another size...

See my comment above regarding thumbnail sizes. We can't generate all possible sizes, so I'm picking one which should work with most desktops: 400x200.
HiDPI screens will ask for a different size, so new thumbs will be generated, but based on the max preview which was generated with the command.

Also, every folder will require 1-4 new thumbnails, depending on how many images it contains.
So, if your top folder contains 50 folders, you will need to generate another 200 thumbnails (max) and if you go inside one and it contains 10 folders, that's another 40 thumbnails.
The best way to check is to capture the logs in order to get the IDs of files you want to test and then to look in their thumbnails folder. Another way is to list your thumbnails folder. Only a few folders should be marked as recently updated when you browse your whole collection.

mmattel:
Here is something new, don´t know in which depth related, lets see.
- deleted all thumbs and upgraded my local test env to your latest PR
- started the ./occ command to create thumbnails (some hours ago) - all fine
- selected on my iPad from my pictures one and tried it to upload via the owncloud client to my server
  scince then I can not access my owncloud test instance anymore, not with the sync client and not with the browser, the latter tells "internal server error".
- the ./occ command continued to run and produced thumbs...

the owncloud log reports:

```
{"reqId":"oJLNcM\/fDLj9cvXacpOZ","remoteAddr":"w.x.y.z","app":"index","message":"Exception: {\"Exception\":\"Doctrine\\\\DBAL\\\\Exception\\\\DriverException\",\"Message\":\"An exception occurred while executing 'UPDATE `oc_file_locks` SET `lock` = `lock` + 1, `ttl` = ? WHERE `key` = ? AND `lock` >= 0' with params [1443700516, \\\"files\\\\\\\/f734dc428f23a4e02ecd79ea7cb5ff96\\\"]:\\n\\nSQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction\",\"Code\":0,\"Trace\":\"
#0 \\\/xxx\\\/3rdparty\\\/doctrine\\\/dbal\\\/lib\\\/Doctrine\\\/DBAL\\\/DBALException.php(116): Doctrine\\\\DBAL\\\\Driver\\\\AbstractMySQLDriver->convertException('An exception oc...', Object(Doctrine\\\\DBAL\\\\Driver\\\\PDOException))\\
n#1 \\\/xxx\\\/3rdparty\\\/doctrine\\\/dbal\\\/lib\\\/Doctrine\\\/DBAL\\\/Connection.php(996): Doctrine\\\\DBAL\\\\DBALException::driverExceptionDuringQuery(Object(Doctrine\\\\DBAL\\\\Driver\\\\PDOMySql\\\\Driver), Object(Doctrine\\\\DBAL\\\\Driver\\\\PDOException), 'UPDATE `oc_file...', Array)\\
n#2 \\\/xxx\\\/lib\\\/private\\\/db\\\/connection.php(203): Doctrine\\\\DBAL\\\\Connection->executeUpdate('UPDATE `oc_file...', Array, Array)\\
n#3 \\\/xxx\\\/lib\\\/private\\\/lock\\\/dblockingprovider.php(112): OC\\\\DB\\\\Connection->executeUpdate('UPDATE `*PREFIX...', Array)\\
n#4 \\\/xxx\\\/lib\\\/private\\\/files\\\/storage\\\/common.php(638): OC\\\\Lock\\\\DBLockingProvider->acquireLock('files\\\/f734dc428...', 1)\\
n#5 \\\/xxx\\\/lib\\\/private\\\/files\\\/storage\\\/wrapper\\\/wrapper.php(571): OC\\\\Files\\\\Storage\\\\Common->acquireLock('files', 1, Object(OC\\\\Lock\\\\DBLockingProvider))\\
n#6 \\\/xxx\\\/lib\\\/private\\\/files\\\/storage\\\/wrapper\\\/wrapper.php(571): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Wrapper->acquireLock('files', 1, Object(OC\\\\Lock\\\\DBLockingProvider))\\
n#7 \\\/xxx\\\/lib\\\/private\\\/files\\\/view.php(1779): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Wrapper->acquireLock('files', 1, Object(OC\\\\Lock\\\\DBLockingProvider))\\
n#8 \\\/xxx\\\/lib\\\/private\\\/files\\\/view.php(1874): OC\\\\Files\\\\View->lockPath('\\\/', 1, false)\\
n#9 \\\/xxx\\\/lib\\\/private\\\/files\\\/view.php(1190): OC\\\\Files\\\\View->lockFile('\\\/', 1)\\
n#10 \\\/xxx\\\/lib\\\/private\\\/files\\\/filesystem.php(861): OC\\\\Files\\\\View->getFileInfo('\\\/', false)\\
n#11 \\\/xxx\\\/apps\\\/files\\\/index.php(91): OC\\\\Files\\\\Filesystem::getFileInfo('\\\/', false)\\
n#12 \\\/xxx\\\/lib\\\/private\\\/route\\\/route.php(154) : runtime-created function(1): require_once('\\\/xxx\\\/w...')\\
n#13 [internal function]: __lambda_func(Array)\\
n#14 \\\/xxx\\\/lib\\\/private\\\/route\\\/router.php(281): call_user_func('?lambda_154', Array)\\
n#15 \\\/xxx\\\/lib\\\/base.php(829):OC\\\\Route\\\\Router->match('\\\/apps\\\/files')\\
n#16 \\\/xxx\\\/index.php(40): OC::handleRequest()\\n#17 {main}\",\"File\":\"\\\/xxx\\\/3rdparty\\\/doctrine\\\/dbal\\\/lib\\\/Doctrine\\\/DBAL\\\/Driver\\\/AbstractMySQLDriver.php\",\"Line\":115}","level":3,"time":"2015-10-01T10:56:07+00:00"}
```

I interrupted the commad (ctrl-c), one core continued to show show 100% cpu with an mysql task which was for sure causd by the lock. I can not get rid of the mysql lock because when I kill the mysql PID´s and start mysql, it immediately comes up with 100% and beeing locked. Also nothing changes restarting nginx and php5-fpm. I also can´t access the db with phpmyadmin... The last time I had that I needed to reinstall my test instance.

mysql> SHOW FULL PROCESSLIST;
image

Could be a problem running the thumbnail generator and running in such an error by concurrently accessing the instance with the sync client - even I think this is not command related, but it raises the problem

Post edit: after about 10min the process ID with the mysql query disappeared, but whatever I do, the database stays locked...

oparoz:
I ran into this before:
https://github.com/owncloud/core/issues/19457

and I think there is a serious problem with the locking mechanism because it can be sent in a infinite loop under certain conditions.

In your case, it's probably caused by the simultaneous access of one file by both the client and the command, but this situation could occur when multiple users are accessing the same folder.

Try using `files:scan` to see if you get your system back.
There must be a running processes telling mySQL to perform some task. If it's not in the list of OS processes, then take down the instance and clear the oc_file_locks table.

I thought about enabling single-user mode when scanning to prevent concurrent access, but then thought that it would be a problem for some users to take down their entire cloud for 1 or more hours.

mmattel:
I have used `files:scan` and it worked, but honestly I have forgotten to check in advance if the cpu load was still up. So I can´t tell if the command worked or not or it just was the time passing by to release the lock (I was out for some hours...)

The idea to go into single user mode is imho a bad idea, sorry. I have a performant cifs server and with concurrent access, it can closely saturate the GbE wire. But looking what the I/O is, the oC server has some load and the cifs server is ideling. Coming back to the quantity of 19K picture files in high res, this takes some time to get processed. Compared to NFS, SMB is slow - I know...

The root cause has to be fixed, such an error is an no go for any production, just my 2c

mmattel:
@oparoz suggestion 
`$ sudo -uwww-data ./occ gallery:create-thumbnails --path=/admin/files/mounted/SMB/`
-->
`$ sudo -uwww-data ./occ gallery:create-thumbnails --path=/admin/files/mount_name`

oparoz:
It's now possible to force the re-generation of thumbnails, in case something went bad in a folder.

> The root cause has to be fixed, such an error is an no go for any production, just my 2c

I agree, even if I was scanning to early, per example, in my early tries, the locking app should not end up in a deadlock.

mmattel:
Just FYI, there are also ~4600 RAW images  

Most of the thumbs have already been created from an earlier run, so it did not start from scratch.
image

oparoz:
Thanks.

Depending on the machine, RAW conversion can take a long time, but it seems scanning that storage location is taking at least an hour.

Once I release, the next version, it would be interesting to compare a normal scan vs a gallery scan (where all thumbs are already generated) to see if there are any bottlenecks.

I might make a slower scan option available, using transactional locking, in order to improve reliability.

oparoz:
Added a delete command. Can also work solely on the cache in case it's out of sync. A wrong cache entry can block the generation of new thumbnails.

oparoz:
Moved the documentation to the wiki

oparoz:
I've looked at the way the cache scanner works and the entire scanned folder is locked and cannot be read by the WebGUI and probably WebDAV while it's being processed.

So the current workflow is
1. Start the DB collector
2. Lock the folder
3. Start scanning the folder recursively
4. Start generating previews as file events arrive
5. Unlock the folder
6. Commit detected changes to the DB
7. Finish generating previews

In my tests, I cannot access the external share while it's being scanned. That could explain the timeouts as I'm sure the GUI will receive that message if it can't access files after x minutes.

In that scenario, it won't help to commit changes to the DB at every step of the way. The best way forward would be to collect structure data first and to launch non-recursive scans from the bottom of the tree. Each folder update would be committed separately.

mmattel:
FYI
- If I mount the picture share (SMB) for the very first time and the scanner goes over it, it scans 18,8K files and 504 directories in less than a minute. This is absolutely ok.
- When the thumnail creation command runs on this mount for the first time, it takes (RAW files are NOT processed) multiple hours (I remember something like 5-6h...). It is shorter but even hours, when the thumbs have been created already (I can give this a try over night, thanks to the result summary :smile:) 
- imho, the current file or at least the current folder to be processed has to be locked and not the whole mount

oparoz:
> When the thumnail creation command runs on this mount for the first time, it takes (RAW files are NOT processed) multiple hours (I remember something like 5-6h...). It is shorter but even hours, when the thumbs have been created already (I can give this a try over night, thanks to the result summary :smile:)

I've just tried with the latest version and scanning is reasonably fast

| Folders | Files | Supported images | New previews | Failed previews | Elapsed time | Total size |
| :-: | :-: | :-: | :-: | :-: | :-: | :-: |
| 15 | 2397 | 1585 | 2 | 1 | 00:00:26.1 | 138.38M |

But, this has not been unmounted before launching the command. I have no idea how to do that as the panel doesn't offer such an option.

It would be interesting to see if by scanning first, the create-thumbnails command is given a boost.

> imho, the current file or at least the current folder to be processed has to be locked and not the whole mount

I agree, but there isn't such an option available, so it will be a while before this gets implemented as it needs to be custom written.

```
customScanner() {
    getDirectoryContent()
    foreach files {cache->scanFile()}
    foreach dirs{customScanner(dir)}
}
```

mmattel:
@oparoz Note: your statistics output would be great to have also in ./occ files:scan ect... :smile:

mmattel:
Info: You can not use `./occ gallery:create-thumbnails --all --quiet` because it asks you 
`Are you sure you want to delete the thumbnails for all users?` but this is not shown on the screen when using `--quiet` ...

This question must be asked, shown and answered and then quiet should take pace

oparoz:
> Note: your statistics output would be great to have also in ./occ files:scan ect... :smile:

I agree. Maybe add an issue in `core` if someone wants to pick up that task?

>  it asks you 
> Are you sure you want to delete the thumbnails for all users?

That's a bug actually. I goofed up the implementation of the questions... The should not be in the super class...

mmattel:
@oparoz 
Related to: Remove oc_storages entries after removing ext storages https://github.com/owncloud/core/pull/19610, merged

Would it be possible to run an automated delete of all thumbs created when a mount is removed?
As the mount with its files is then no longer present, any thumbs created are "orphans"...
(When a mount is removed, it naturally affects all users)

mmattel:
Info:
I have run thumbnail creation sucessfully.
Then I wanted to see how long it takes running the command again. Note one picture is bad.
I ran the command, once with --quiet and once without.
Running with --quiet, I do get a php error message printed on the screen and no final statistics, running it without quiet, I do not get the php error printed on the screen but as expected the statistics:

```
sudo –u www-data ./occ gallery:create-thumbnails --all –quiet

PHP Parse error:  imagecreatefromjpeg(): gd-jpeg, libjpeg: recoverable error:
 in /xxxx/lib/private/image.php on line 512
PHP Stack trace:
PHP   1. {main}() /xxxx/occ:0
PHP   2. require_once() /xxxx/occ:11
PHP   3. OC\Console\Application->run() /xxxx/console.php:69
PHP   4. Symfony\Component\Console\Application->run() /xxxx/lib/private/console/application.php:97
PHP   5. Symfony\Component\Console\Application->doRun() /xxxx/3rdparty/symfony/console/Symfony/Component/Console/Application.php:126
PHP   6. Symfony\Component\Console\Application->doRunCommand() /xxxx/3rdparty/symfony/console/Symfony/Component/Console/Application.php:195
PHP   7. Symfony\Component\Console\Command\Command->run() /xxxx/3rdparty/symfony/console/Symfony/Component/Console/Application.php:874
PHP   8. OCA\Gallery\Command\CreateThumbnails->execute() /xxxx/3rdparty/symfony/console/Symfony/Component/Console/Command/Command.php:253
PHP   9. OCA\Gallery\Command\CreateThumbnails->scanAndGenerateMissingThumbnails() /xxxx/apps/gallery/command/createthumbnails.php:224
PHP  10. OCA\Gallery\Command\CreateThumbnails->scanForMissingThumbnails() /xxxx/apps/gallery/command/createthumbnails.php:310
PHP  11. OCA\Gallery\Command\Thumbnails->performOperation() /xxxx/apps/gallery/command/createthumbnails.php:334
PHP  12. call_user_func_array() /xxxx/apps/gallery/command/thumbnails.php:297
PHP  13. OCA\Gallery\Command\CreateThumbnails->scanAndUpdateCache() /xxxx/apps/gallery/command/thumbnails.php:297
PHP  14. OCA\Gallery\Command\GalleryScanner->scan() /xxxx/apps/gallery/command/createthumbnails.php:287
PHP  15. OC\Files\Cache\Scanner->scan() /xxxx/apps/gallery/command/galleryscanner.php:95
PHP  16. OC\Files\Cache\Scanner->scanChildren() /xxxx/lib/private/files/cache/scanner.php:268
PHP  17. OC\Files\Cache\Scanner->scanChildren() /xxxx/lib/private/files/cache/scanner.php:385
PHP  18. OC\Files\Cache\Scanner->scanChildren() /xxxx/lib/private/files/cache/scanner.php:385
PHP  19. OC\Files\Cache\Scanner->scanFile() /xxxx/lib/private/files/cache/scanner.php:345
PHP  20. OC\Hooks\BasicEmitter->emit() /xxxx/lib/private/files/cache/scanner.php:191
PHP  21. call_user_func_array() /xxxx/lib/private/hooks/emittertrait.php:98
PHP  22. OCA\Gallery\Command\GalleryScanner->OCA\Gallery\Command\{closure}() /xxxx/lib/private/hooks/emittertrait.php:98
PHP  23. call_user_func_array() /xxxx/apps/gallery/command/galleryscanner.php:128
PHP  24. OCA\Gallery\Command\CreateThumbnails->generateMissingThumbnail() /xxxx/apps/gallery/command/galleryscanner.php:128
PHP  25. OCA\Gallery\Command\CreateThumbnails->createThumbnailForSupportedMediaType() /xxxx/apps/gallery/command/createthumbnails.php:124
PHP  26. OCA\Gallery\Command\CreateThumbnails->checkCacheAndCreateThumbnail() /xxxx/apps/gallery/command/createthumbnails.php:391
PHP  27. OCA\Gallery\Command\CreateThumbnails->CreateThumbnail() /xxxx/apps/gallery/command/createthumbnails.php:434
PHP  28. OC\Preview->getPreview() /xxxx/apps/gallery/command/createthumbnails.php:455
PHP  29. OC\Preview->generatePreview() /xxxx/lib/private/preview.php:770
PHP  30. OC\Preview\Image->getThumbnail() /xxxx/lib/private/preview.php:1136
PHP  31. OC_Image->loadFromFile() /xxxx/lib/private/preview/image.php:55
PHP  32. imagecreatefromjpeg() /xxxx/lib/private/image.php:512
```

oparoz:
> Would it be possible to run an automated delete of all thumbs created when a mount is removed?

Yes, but that's a job for the Preview class. I wonder if it doesn't do it already since the cache must be firing hooks when remove items.

> Running with --quiet, I do get a php error message printed on the screen and no final statistics

I can see why that would happen. The error catching is too intertwined with the quiet mode. I'll take a look next time I open the project.

oparoz:
@mmattel - Did you still experience 1205 errors?

budulinek:
Hi oparoz,
thanks a lot! I tried to run thumbcreation. 

sudo -uowncloud php-cli ./occ gallery:create-thumbnails --all

First three files were processed without problems, but then I ran into 
Segmentation fault

I have restarted the command and again, Segmentation fault after 2 processed files.

strace gave me this: 

..........
write(4, "Preview generated!\n", 19Preview generated!
)    = 19
munmap(0x4fd5a000, 266240)              = 0
munmap(0x4fcd8000, 266240)              = 0
.........
munmap(0x4a170000, 266240)              = 0
munmap(0x4a12f000, 266240)              = 0
munmap(0x4a0ee000, 266240)              = 0
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x200c} ---
+++ killed by SIGSEGV +++
Segmentation fault

I know nothing about PHP or programming, so I have no idea what could be the source of the problem. Maybe some issue with memory management on my side. I tried to turn off APC memcache, but it did not help.

oparoz:
@budulinek - Maybe a missing library if you compile stuff yourself? In any case, I would disable all providers and enable them one by one to see if there is a problem with a specific one. Also, it seems the BMP writer in `core` exhausts PHP's memory, so if you have some of these, try to move them away from the area you're trying to scan.

oparoz:
The Scanner was fixed in `core` for people using the database as a locking backend, so I've ported the fix over.

With this, I am able to browse files as the previews are generated by the command and new previews generated while browsing are recognised as already existing by the scanner: win-win

Hopefully, it will work as well for you @mmattel and other users of the command.

oparoz:
@mmattel - Regarding the PHP error you get, you need this: https://github.com/owncloud/gallery/pull/422

And regarding the quiet mode, it turns out it wasn't entirely my fault. It is managed by Symfony at a global level and it prevents their question helper from printing questions... at least in the older version of the component shipping in `core`.
I've fixed the stats so that they print, no matter what.

command/thumbnails.php
```diff
@@ -0,0 +1,473 @@
...
+	 * @param OutputInterface $output
+	 *
+	 * @throws InputCommandException
+	 */
+	protected function getConfirmation(InputInterface $input, $question, OutputInterface $output) {
+		$helper = $this->getHelper('question');
+		$confirmationQuestion = new ConfirmationQuestion($question, false);
+
+		$quiet = false;
+		if ($output->isQuiet()) {
```
nickvergessen: command/thumbnail.php

bildschirmfoto vom 2015-10-14 13 23 29

command/thumbnails.php
```diff
@@ -0,0 +1,473 @@
...
+			$quiet = true;
+			$output->setVerbosity($output::VERBOSITY_NORMAL);
```
nickvergessen: if you change the `$output::VERBOSITY_*` stuff to `OutputInterface::VERBOSITY_*` it solves the issue with the check-code

command/thumbnails.php
```diff
@@ -0,0 +1,473 @@
...
+		}
+		if (!$helper->ask($input, $output, $confirmationQuestion)) {
+			throw new InputCommandException("Operation aborted");
+		}
+		if ($quiet) {
+			$output->setVerbosity($output::VERBOSITY_QUIET);
```
nickvergessen: if you change the `$output::VERBOSITY_*` stuff to `OutputInterface::VERBOSITY_*` it solves the issue with the check-code

command/thumbnails.php
```diff
@@ -0,0 +1,473 @@
...
+	/**
+	 * Presents the results of the operation in tabular form
+	 *
+	 * @param OutputInterface $output
+	 */
+	protected function presentResults(OutputInterface $output) {
+		// Stop the timer
+		$this->execTime += microtime(true);
+
+		$output->setVerbosity($output::VERBOSITY_NORMAL);
```
nickvergessen: if you change the `$output::VERBOSITY_*` stuff to `OutputInterface::VERBOSITY_*` it solves the issue with the check-code

command/thumbnails.php
oparoz: My EAP version of PHPStorm is not working (I'm having loads of false positives) and I forgot to set the branch when looking at the class online... It's not even in 2.8, only in master... 

Will use `getVerbosity`

js94x:
:+1:

jospoortvliet:
Time to merge? ;-)

oparoz:
I know it would help a lot of people, but I (or any volunteer really) need to write all the tests for the new commands.

mmattel:
@oparoz 
FYI, the statistics output for occ files:scan which I based on your code for statistics from here, has been merged today :+1: https://github.com/owncloud/core/pull/21201

This more reminds us to fix https://github.com/owncloud/core/issues/19458 "Scanner sends folder through the scanFile event"

jospoortvliet:
I just applied the patch here locally and ran the command with --all. I was tired of waiting for thumbnail generation all the time ;-)

The command returns right away, giving no output at all. Clearly, thumbs aren't generated, which I can see when visiting a folder I haven't visited before.

Log output contains:
`Error: Call to undefined function OCA\Gallery\Command\pcntl_signal() at /srv/www/htdocs/owncloud/apps/gallery/command/thumbnails.php#138`

This has been applied on gallery in ownCloud 8.2.2.

Patching went without issue, I ran it this way::

```
$ patch -p1 < 82599da05676617b8f242203bdfdf8f67b0b2ed4.patch
patching file appinfo/register_command.php
patching file command/commandexception.php
patching file command/createthumbnails.php
patching file command/deletethumbnails.php
patching file command/forbiddencommandexception.php
patching file command/galleryscanner.php
patching file command/inputcommandexception.php
patching file command/notfoundcommandexception.php
patching file command/thumbnails.php
```

I reversed the patch for now.

oparoz:
That's not a proper bug report!!! ;)

PHP version please. I suspect you're using 7 :)

LukasReschke:
pcntl_signal is in the PCNTL extension which is not installed by default with PHP. A lot of environment even miss it and in some cases it even needs to be compiled manually.

oparoz:
Ah, thanks. I knew it had to be added to the PHP config, but thought the bug was about a namespace issue...
I guess I'm going to need to add a few IFDEFs

jospoortvliet:
I did `zypper in php7-pcntl`, restarted and got this error upon running with the --all command:

```
[OCP\Files\StorageNotAvailableException]
```

Next, he ran with his own username and the system was busy for 'a while' (a looooong while). Feedback (that something is happening) would be nice ;-)

Lots of error messages including:

```
PHP Warning:  Uncaught ErrorException: imagecreatefrompng(): gd-png: fatal libpng error: IDAT: CRC error
 in Unknown:0
Stack trace:
#0 [internal function]: OCA\Gallery\Command\Thumbnails->exceptionErrorHandler(3, 'imagecreatefrom...', 'Unknown', 0, Array)
#1 /srv/www/htdocs/owncloud/lib/private/image.php(519): imagecreatefrompng('/srv/www/htdocs...')
#2 /srv/www/htdocs/owncloud/lib/private/preview/image.php(55): OC_Image->loadFromFile('/srv/www/htdocs...')
#3 /srv/www/htdocs/owncloud/lib/private/preview.php(1135): OC\Preview\Image->getThumbnail('XXX/XXXXX/...', 2048, 2048, false, Object(OC\Files\View))
#4 /srv/www/htdocs/owncloud/lib/private/preview.php(770): OC\Preview->generatePreview(84687)
#5 /srv/www/htdocs/owncloud/apps/gallery/command/createthumbnails.php(443): OC\Preview->getPreview()
#6 /srv/www/htdocs/owncloud/apps/gallery/command/createthumbnails.php(423): OCA\Gallery\Command\CreateThumbnails->CreateThumbnail(Object(OC\Preview), '/XXXXXX/...', Object(Symfony\Component\Console\Output\ConsoleOutput))
# in Unknown on line 0
PHP Parse error:  imagecreatefrompng(): gd-png error: setjmp returns error condition 2
 in /srv/www/htdocs/owncloud/lib/private/image.php on line 519
```

and

```
libpng warning: iCCP: cHRM chunk does not match sRGB
libpng warning: iCCP: too many profiles
libpng warning: iCCP: known incorrect sRGB profile'
libpng warning: iTXt: CRC error
libpng warning: Interlace handling should be turned on when using png_read_image
```

(I guess those are just about broken images...)

But I can confirm that it works now: thumbnails seem to have been generated. Yay!

Merge it ;-)

Oh, the tests :disappointed:

oparoz:
> Next, he ran with his own username and the system was busy for 'a while' (a looooong while). Feedback (that something is happening) would be nice ;-)

I'm confused, because something should show up as soon as a file is found and you're scanning the **file**system.

> Lots of error messages 

Yes, but the nice thing is that it doesn't kill the process ;). In verbose mode, you should even see which images are broken.

> Oh, the tests :disappointed:

Yes, that and more real world testing needed on various platforms

elpraga:
That is really great news!
If I understand it correctly, right now it generates only preview_max, is that so? (I will have to wait until 9.0 is released to try it out myself)

Would it be difficult to make it generate some more thumbnails from preview_max giving it for example one dimension? (I know there are no standard values for previews yet, :( ... )

oparoz:
> If I understand it correctly, right now it generates only preview_max, is that so?

Correct

> I will have to wait until 9.0 is released to try it out myself

No need to wait, just load a VM of 9.0 and if it doesn't exist, just load the current one and clone `core`master (easy and documented)

> Would it be difficult to make it generate some more thumbnails from preview_max giving it for example one dimension?

No, I'd love for this feature to be available, so don't hesitate to submit a patch :)

jospoortvliet:
@oparoz fixed my comment formatting so it actually shows some stuff that was hidden ;-)

Pls re-read!

oparoz:
Regarding `[OCP\Files\StorageNotAvailableException]`, you should get the same error if you run a simple scanfile. There must be a user without storage, maybe a test account which gets wiped during unit testing or a user using mounted storage.

Does it completely stop the command? If yes, then I should add something to simply skip that user.

Apart from that, how is the speed to get all your thumbnails? On what hardware? And with how large a sample?

jospoortvliet:
No, it didn't stop anything.

It took some hours, I believe, but then again I have gigabytes of pictures!

oparoz:
I'm amazed that it still took hours if that was on an instance where you have Gallery installed as some thumbnails must have been generated, but I guess it's still a valid option for some people who want to mount large media collections to their ownCloud.

mmattel:
@oparoz as you see by the comments, your code is very important to the public and just waits to get merged :+1:  :smile:

oparoz:
I think it will be even more important for underpowered devices, but we really need the tests in so that we can add more features.

elpraga:
I´d love to be able to submit a patch! Unfortunately I am not :( 

The improvement You made is going to help a lot as it is already, so big thanks for the new command!! :)

mmattel:
@oparoz 
Just did the following:
- cloned a new test instance via ocdev setup base and finalized setup
- applied this PR on it 
- try to run ./occ

```
yyyy:/xxxx/apps/gallery$ sudo git pr 387
From https://github.com/owncloud/gallery
 * [new ref]         refs/pull/387/head -> pr/387
Switched to branch 'pr/387'
yyyy:/xxxx/apps/gallery$ cd ../..
yyyy:/xxxx$ sudo -uwww-data ./occ
PHP Fatal error:  Class 'OC\Files\Cache\ChangePropagator' not found in /xxxx/apps/gallery/command/createthumbnails.php on line 93
PHP Stack trace:
PHP   1. {main}() /xxxx/oclg/occ:0
PHP   2. require_once() /xxxx/oclg/occ:11
PHP   3. OC\Console\Application->loadCommands() /xxxx/oclg/console.php:80
PHP   4. require() /xxxx/oclg/lib/private/console/application.php:73
PHP   5. OC\AppFramework\Utility\SimpleContainer->query() /xxxx/oclg/apps/gallery/appinfo/register_command.php:46
PHP   6. OC\AppFramework\Utility\SimpleContainer->resolve() /xxxx/oclg/lib/private/appframework/utility/simplecontainer.php:104
PHP   7. OC\AppFramework\Utility\SimpleContainer->buildClass() /xxxx/oclg/lib/private/appframework/utility/simplecontainer.php:83
PHP   8. ReflectionClass->newInstanceArgs() /xxxx/oclg/lib/private/appframework/utility/simplecontainer.php:66
PHP   9. OCA\Gallery\Command\CreateThumbnails->__construct() /xxxx/oclg/lib/private/appframework/utility/simplecontainer.php:66
```

oparoz:
Thanks for the report. Looks like I'm going to have to create 2 different versions. One for 9 and one for 8.2

davodego:
Great idea :) 
I installed the patch -- all well, no errors. But when I run "occ gallery:create-thumbnails --all"  occ says: " [InvalidArgumentException]
  There are no commands defined in the "gallery" namespace."

What can I do?

oparoz:
@davodego - Check the file permissions

davodego:
@oparoz permissions of which file? I have installed the patch as web user www-data.

oparoz:
That's the one's I was thinking about.
If you do `./occ`, the commands should be listed. If they're not, then that means the commands were not registered for some reason.

mmattel:
@oparoz 
**Important note**
I remember that you also added the functionality to interrupt a running occ gallery task by ctrl-c.
As I have this functionality also in my core PR occ scan:files Adding more details in the base print out (II), PVince81 identified during testing an issue that `pcntl_signal` functions are not existing in his environment. (to be enabled somehow, but this is a different story). Please see his report: https://github.com/owncloud/core/pull/21701#issuecomment-173944678 and the following comments.
I had to adopt the code to check if `pcntl_signal` functions are existing or not to avoid PHP errors. You can see my changes in the mentioned core PR. Maybe this if of help for you.

oparoz:
Thanks, [it's been reported before](https://github.com/owncloud/gallery/pull/387#issuecomment-169405501) and was added to the todo list. I'll take a look at how you've worked around it as it might speed things up :)

mmattel:
@oparoz FYI, pls see https://github.com/owncloud/core/pull/22023 (Interruptable commands)
Interruption handling code for occ commands is now part of core. You also see in scan.php how it is used.

oparoz:
I'm glad it's made it to the base class :). Thanks for the heads up.

elpraga:
I just applied this patch on OC 9.0 (using the guide from http://blog.jospoortvliet.com/2016/01/patching-owncloud-get-your-fix-now.html?m=1). 
I checked before applying the patch, the check went OK, I applied the patch with no error reported, but when I try to run the occ command, I get an error on line 93. 

Does somebody know when I'm doing an error?

oparoz:
Cool guide @jospoortvliet :)

@elpraga - The problem is that things have changed in `core` between the last commit in this PR and today, so the patch is sadly out of date and won't give you what you want.

elpraga:
OK. Thanks for explaining! Are the changes big? 

In other words, is there a chance that the branch gets updated and the patch will be applicable to 9.0? 

I was waiting quite anxiously for 9.0 to try out the command for thumbnail generation and I was really sad it did not maker it after all.. :-(

oparoz:
> Are the changes big?

No idea. It could be relatively quick to just fix, but some refactoring will be needed on 9.0 as some of the features developed in here have been integrated in `core`.

It's been low on my list of priorities unfortunately as I can't justify spending 2-3 days to fix it.

elpraga:
Thank you for clarifying! I understand that priorities are tough sometimes. 

Unfortunately I cannot afford to pitch in some money right now nor offer my skills to write the fix as I lack them..

I'll have to wait then.. But thank you a lot anyways!

Bugsbane:
Hoping that this will live to see OC 9+

zbirmingham0001:
Agreed, Love to see this working with OC9. This is why I use OC but the thumbnail generation is holding us back.

oparoz:
Closing as this will not be added to Gallery

elpraga:
So this functionality will not be added not even to 9.1? Really? That is really sad to hear, I was waiting for it for months...
Is there any chance for the decision to be reversed?

jospoortvliet:
@elpraga if somebody writes the tests, I suppose ;-)

elpraga:
Unfortunately, I don't know how to. Otherwise, I would offer to help.. :-/

oparoz:
Write tests, remove code made obsolete by changes in core, use the background queue if it ever gets documented (and is useful) plus much more I'm sure.

Since I'm going to focus on Gallery+, this may make a come back there.

nksupport:
@oparoz, i'm sure there's a whole lot of us who will be happy with just an ugly hack to run in the shell, something that would call the thumb generation method directly, not integrated in core or anything. ditaW pasted a browser-side greasemonkey script [here](https://github.com/owncloud/core/issues/11685) , but it seems broken already... any hints?

DeepDiver1975:
Form my pov this should go into core ....

oparoz:
@DeepDiver1975 - I agree, all the base components are there. This was sort of a crutch while waiting for some movement on a background task queue manager capable of generating thumbnails.

DeepDiver1975:
> background task queue manager capable of generating thumbnails.

we should start up the discussion about this for the scope of 9.2 again

oparoz:
For reference this is the issue in core: [Migrate preview generation to async job execution](https://github.com/owncloud/core/issues/11685)

PR #391, opened 2015-09-28
Purify SVGs

### LINKED ISSUE #373: Use DOMPurify to parse SVGs [$15]
@LukasReschke - What do you think about using DOMPurify to parse SVGs before showing them in the browser?
- [x] Decode and parse base64 encoded SVG thumbnails
- [x] Parse large SVG previews
## <bountysource-plugin>

Did you help close this issue? Go claim the **[$15 bounty](https://www.bountysource.com/issues/26922629-use-dompurify-to-parse-svgs?utm_campaign=plugin&utm_content=tracker%2F9328526&utm_medium=issues&utm_source=github)** on [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F9328526&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oparoz:
Fixes #373

Some purified SVGs can't be displayed until DOMPurify 0.71 is released: cure53/DOMPurify@86929b9

Good, simple SVGs on the other hand are properly displayed.
### Notes
- The API is not affected by the change since the client has to explicitly ask for a SVG and can implement his own purifier
- Native SVGs are not supported anywhere on IE8 and IE9
- Native SVGS are not supported in Gallery on all version of IE, but you do get a preview in the slideshow, even though you don't get a thumbnail...

@LukasReschke

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -71,6 +71,7 @@ if (isset($request->server['REQUEST_URI'])) {
 		 */
 		Util::addScript($appName, 'vendor/bigshot/bigshot-compressed');
 		Util::addScript($appName, 'vendor/image-scale/image-scale.min');
+		Util::addScript($appName, 'vendor/dompurify/src/purify');
 		Util::addScript($appName, 'galleryfileaction');
 		Util::addScript($appName, 'slideshow');
 		Util::addScript($appName, 'slideshowcontrols');
--- a/js/galleryfileaction.js
+++ b/js/galleryfileaction.js
@@ -37,6 +37,10 @@
 		register: function (mediaTypes) {
 			//console.log("enabledPreviewProviders: ", mediaTypes);
 			if (mediaTypes) {
+				// Remove SVG if the user is using an insecure browser (IE8-9)
+				if (window.galleryFileAction.features.indexOf('native_svg') > -1 && !window.btoa) {
+					mediaTypes.splice(mediaTypes.indexOf('image/svg+xml'), 1);
+				}
 				galleryFileAction.mediaTypes = mediaTypes;
 			}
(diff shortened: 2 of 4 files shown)


oparoz:
Rendering fix should be in the next DOMPurify: https://github.com/cure53/DOMPurify/commit/86929b9ec421f941cddb132d5abd7e6340641c5a

Firefox can't show SVGs in the slideshow on the Files side unless this gets merged: 
https://github.com/owncloud/core/pull/19426

oparoz:
@LukasReschke - Could you quickly check that this is the right approach?

### Photowall
- We receive base64 encoded data
- Data is decoded and sanitized
- Data is re-encoded and added to the src property of an Image object

### Slideshow
- We receive a URL
- URL is used in an XMLHttpRequest to retrieve the image
- Data is parsed
- Data is base64 encoded and added to the src property of an Image object

appinfo/app.php
```diff
@@ -71,6 +71,7 @@ function () use ($c, $appName) {
 		 */
 		Util::addScript($appName, 'vendor/bigshot/bigshot-compressed');
 		Util::addScript($appName, 'vendor/image-scale/image-scale.min');
+		Util::addScript($appName, 'vendor/dompurify/src/purify');
```
icewind1991, line 74: Forgot to commit the lib?

oparoz, line 74: It's already there, but wasn't made available in the slideshow before.

icewind1991, line 74: Ok, was looking the wrong vendor folder for it

LukasReschke:
:+1:

oparoz:
Thanks!

PR #394, opened 2015-09-30
domHeight spelling

phil-davis:
I think this is localized to code in this file only, and so does not cause any actual bug.


oparoz:
Luckily, they were all misspelled :). Thanks for the PR.

PR #415, opened 2015-10-09
Fix broken spinner in apps menu

### LINKED ISSUE #19673: App menu spinner shifted
### Steps to reproduce

1. Go to the files app (important!)
1. Navigate to the apps page or any other app using the app menu on the top left

### Expected result
Aligned spinner

### Actual result
files-appmenu-spinner

### Versions
ownCloud 8.2 RC1

This only happens while the files app is open so I suspect there might be a CSS rule conflict that shifts the spinner there.

@owncloud/designers

MorrisJobke:
- fixes owncloud/core#19673

cc @oparoz @pvince81 

I tested this with the slideshow and the spinner still shows up.

### DIFF
--- a/css/slideshow.css
+++ b/css/slideshow.css
@@ -197,7 +197,7 @@
 	height: 100%;
 }
 
-.icon-loading-dark {
+#slideshow .icon-loading-dark {
 	position: absolute !important;
 	top: 50%;
 	left: 0;


PVince81:
Tested, works :+1:

oparoz:
Thanks for the fix. I need to check if that still works for themes.

oparoz:
:+1:

PR #421, opened 2015-10-11
Simplify and fix album thumbnail presentation

### LINKED ISSUE #370: Old folder icons shown when folder is empty (or only has erroneous images and a subfolder in it)
capture du 2015-09-23 22-28-53

Please use the .icon-folder from core – directly reference `core/img/filetypes/folder.svg` instead of duplicating the icon.

### LINKED ISSUE #389: Slideshow notifications are now too low
The top menu is pushing the element down. We need to absolutely position it.

### LINKED ISSUE #390: Hide filename when moving to next picture
The next one may fail and we don't want to see the filename of the previous one.

### LINKED ISSUE #407: Pause button is sometimes not clickable
During a slideshow the pause button in the bottom right is sometimes not clickable.
I believe this happens for every image after it is displayed half of the time.
This issue occurs in Chrome and Firefox.

edit:
this happens because the progress indicator is not clickable.
`<div class="progress icon-view-pause" style="height: 23.72px; overflow: hidden;"></div>`
A simple fix would be to set `pointer-event: none` for this element, but this isn't supported in IE10 and below (http://caniuse.com/#feat=pointer-events).

oparoz:
### Main change

I've simplified the way album thumbnails are checked, discarded and put together.
There should now be an album icon in case of major failure when retrieving thumbnails belonging to that sub-album.
Commits:
- https://github.com/owncloud/gallery/commit/72d901dce19309f070d8e3471c99339cc1fb2587
- https://github.com/owncloud/gallery/commit/a873e1792b26b0cd1f7e052911294cf86dab38d1
### Minor changes

At the same time I've fixed some minor issues and kept that in separate commits.
- Fixes #390
- Fixes #389
- Fixes #407 (already approved)
- Better fix for #370
- Removed obsolete IE CSS

### DIFF
--- a/css/gallerybutton.css
+++ b/css/gallerybutton.css
@@ -14,7 +14,6 @@
 }
 
 #controls .button.view-switcher img{
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=30)";
 	filter: alpha(opacity=30);
 	opacity: .3;
 }
@@ -27,7 +26,6 @@
 }
 
 #controls .button.view-switcher.gallery img{
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=50)";
 	filter: alpha(opacity=50);
 	opacity: .5;
 }
@@ -61,7 +59,6 @@
 .button.left-switch-button img,
 .button.right-switch-button img {
 	vertical-align: -3px;
-	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=60)";
 	filter: alpha(opacity=60);
 	opacity: 0.6;
 }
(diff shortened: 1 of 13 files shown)


jancborchardt:
@oparoz these should be individual pull requests, otherwise it’s very hard to review and there is a big potential for regressions. Some changes will have issues and will block the changes which are fine.

oparoz:
> otherwise it’s very hard to review

I agree, but this is more the next release than a PR because I need to move forward. Most are minor fixes which wouldn't even get a PR.
The main thing to review is that albums are fixed when there are problems with the images found in the folders below: https://github.com/owncloud/gallery/commit/a873e1792b26b0cd1f7e052911294cf86dab38d1

jancborchardt:
Well, still it’s a big chunk of unrelated changes. cc @DeepDiver1975 @Henni @PVince81 @MorrisJobke what’s the call?

PR #440, opened 2015-10-15
Optimize breadcrumbs

oparoz:
Only hide/show elements instead of rebuilding the whole breadcrumbs every time it's being resized
Use Handlebars for the template to make it easier to modify.

The improvements are not really perceptible, but when you're resizing large photowall every little bit helps.
- [x] Breadcrumb works when navigating through albums
- [x] Breadcrumb shrinks when reaching deep albums
- [x] Ellipsis allows navigation to parent folder
- [x] Tooltip shows parent path
- [x] Breadcrumbs is usable on empty albums

_Note: includes #421 and #435_

---

@setnes @demattin if you want to give it a go

### DIFF
--- a/js/breadcrumb.js
+++ b/js/breadcrumb.js
@@ -1,117 +1,114 @@
-/* global Gallery */
+/* global Handlebars, Gallery */
 (function ($, OC, t, Gallery) {
 	"use strict";
+
+	var TEMPLATE =
+		'{{#each crumbs}}' +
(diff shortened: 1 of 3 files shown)


oparoz:
There is a bug which kills the app. I need to find a better way to prevent recalculating when not necessary.

oparoz:
Bug was fixed on the 16th.

PR #442, opened 2015-10-18
create specific folder icon for Gallery, fix #433

### LINKED ISSUE #433: Create gallery album icon for when we can't show images
When we can't fetch at least one thumbnail to put in the album representation, we get a folder icon, but I'd like to have a proper album icon, based on something like this

---

@jancborchardt

jancborchardt:
Fix #433 by adding a specific Gallery-folder-icon. Please review @oparoz @owncloud/designers


oparoz:
Maybe put it in core so that people can theme it like the rest of the icons? At some point, you could have albums for music, documents etc and people could pick the folder they want to use in the Files app in case it requires too much power to detect automatically.

mmattel:
:+1:

oparoz:
@jancborchardt What do you think about having all sorts of multimedia folder icons in core for easy theming?

jancborchardt:
@oparoz I’d say we should first get this simple change in and improve later. :) For now it’s just a file type icon change.

PR #453, opened 2015-10-27
[+] Exif Description

0xb0ba:
show exif description (if exists) instead filename on preview and slideshow
## TODO

Designers need to figure out how to properly show that description. 
- Centred? 
- 3 lines? In this example it would just be the first line over 3 narrower lines
- Scroller or rollover? Rollover would fill the screen with that description since we have to keep it narrow. Scroller requires 2 additional touch areas instead of an ugly scrollbar

A lot of galleries show the exif data in a side-bar.

default

default

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -96,6 +96,12 @@ return [
 			'url'  => '/preview/{fileId}',
 			'verb' => 'GET'
 		],
+		// exif of picture
+                [
+                        'name' => 'files#exif',
+                        'url'  => '/files/exif',
+                        'verb' => 'GET'
+                ],
 		/**
 		 * Public services
 		 */
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -99,7 +99,7 @@ return [
 		// exif of picture
                 [
                         'name' => 'files#exif',
-                        'url'  => '/files/exif',
+                        'url'  => '/files/exif/{fileId}',
                         'verb' => 'GET'
                 ],
 		/**
@@ -131,6 +131,12 @@ return [
 			'url'  => '/preview.public/{fileId}',
 			'verb' => 'GET'
 		],
+		// exif of picture
+                [
+                        'name' => 'files_public#exif',
+                        'url'  => '/files.public/exif/{fileId}',
+                        'verb' => 'GET'
+                ],
 		/**
 		 * API
 		 */
--- a/controller/files.php
+++ b/controller/files.php
@@ -53,21 +53,22 @@ trait Files {
          * @param string $location a path picture
          * @return array
          */
-        public function exif($location){
-            $filename=\OC\Files\Filesystem::getLocalFile($location);
-            $exif = false;
+        public function exif($fileId){
+	    $path=\OC\Files\Filesystem::getPath($fileId);
+	    $filename=\OC\Files\Filesystem::getLocalFile($path);
+	    $exif = false;
 	    $iptc = false;
 	    if (is_callable('exif_read_data')) {
 		$exif=@exif_read_data($filename);
 	    }
-            getimagesize($filename,$info);
-            if (is_array($info)){
-		foreach($info as $k => $i){
-		    $iptc[$k]=iptcparse($i);
+	    getimagesize($filename,$blocks);
+	    if (is_array($blocks)){
+		foreach($blocks as $key => $block){
+		    $iptc[$key]=iptcparse($block);
 		}
 	    }
-            return ['exif'=>$exif,'iptc'=>$iptc];
-        }
+	    return ['exif'=>$exif,'iptc'=>$iptc];
+	}
 
 	/**
 	 * @NoAdminRequired
--- a/controller/files.php
+++ b/controller/files.php
@@ -54,8 +54,10 @@ trait Files {
          * @return array
          */
         public function exif($fileId){
-	    $path=\OC\Files\Filesystem::getPath($fileId);
-	    $filename=\OC\Files\Filesystem::getLocalFile($path);
+	    $file = $this->downloadService->getResourceFromId($fileId);
+	    $path=$file->getInternalPath();
+	    $storage=$file->getStorage();
+	    $filename=$storage->getLocalFile($path);
 	    $exif = false;
 	    $iptc = false;
 	    if (is_callable('exif_read_data')) {
--- a/controller/files.php
+++ b/controller/files.php
@@ -45,6 +45,30 @@ trait Files {
 	/** @var ILogger */
 	private $logger;
 
+        /**
+         * @NoAdminRequired
+         *
+         * Returns a exif of picture
+         *
+         * @param string $location a path picture
+         * @return array
+         */
+        public function exif($location){
+            $filename=\OC\Files\Filesystem::getLocalFile($location);
+            $exif = false;
+	    $iptc = false;
+	    if (is_callable('exif_read_data')) {
+		$exif=@exif_read_data($filename);
+	    }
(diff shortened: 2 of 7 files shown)


oparoz:
Thanks!
We'll first need to discuss if that's something we want to add, but in the meantime, I'll just write down notes about the changes needed before this can be merged.

@jancborchardt and his team can check the design.

appinfo/routes.php
```diff
@@ -96,6 +96,12 @@
 			'url'  => '/preview/{fileId}',
 			'verb' => 'GET'
 		],
+		// exif of picture
+                [
+                        'name' => 'files#exif',
+                        'url'  => '/files/exif',
+                        'verb' => 'GET'
+                ],
 		/**
```
oparoz, line 105: The public routes will need the same endpoint

controller/files.php
```diff
@@ -45,6 +45,30 @@
...
 
+        /**
+         * @NoAdminRequired
+         *
+         * Returns a exif of picture
+         *
+         * @param string $location a path picture
+         * @return array
+         */
+        public function exif($location){
```
oparoz: It's best to use the file's ID

controller/files.php
```diff
@@ -45,6 +45,30 @@
...
+            $filename=\OC\Files\Filesystem::getLocalFile($location);
```
oparoz: We cannot use private APIs

controller/files.php
```diff
@@ -45,6 +45,30 @@
...
+            $exif = false;
+	    $iptc = false;
+	    if (is_callable('exif_read_data')) {
+		$exif=@exif_read_data($filename);
```
oparoz, line 64: Does that work with encrypted files or files located on external storage?

js/slideshow.js
```diff
@@ -103,6 +103,38 @@
 
 				// check if we moved along while we were loading
 				if (currentImageId === index) {
+                                        // check if not in cache
+                                        if (this.images[index].desc===undefined){
```
oparoz, line 108: We need to make sure we only send the request when we have a media type which supports meta data

js/slideshow.js
```diff
@@ -103,6 +103,38 @@
...
+                                            var params={location:this.images[index].path}
+                                            var url=OC.generateUrl('apps/gallery/files/exif?location={location}', params);
```
oparoz: It's best to use something like this `Gallery.utility.buildGalleryUrl('files', '/exif', params);`

js/slideshow.js
```diff
@@ -103,6 +103,38 @@
...
+                                            $.getJSON(url).then(function(data){
+                                                console.log(data);
```
oparoz: To be removed at the end of the process

js/slideshow.js
```diff
@@ -103,6 +103,38 @@
...
+                                                var d;
```
oparoz: I would prefer it if the variable is more descriptive

js/slideshow.js
0xb0ba: buildGalleryUrl is defined in galleryutility.js, but its not included in /apps/files.

oparoz: Ah, yes, you're correct. I keep having to duplicate methods because we don't load too many files on the files side.

appinfo/routes.php
0xb0ba, line 105: done

controller/files.php
0xb0ba: done

controller/files.php
0xb0ba, line 64: yes. for external "local" return direct path, for external nonlocal return path to cached file.

js/slideshow.js
0xb0ba, line 108: done

js/slideshow.js
0xb0ba: done

js/slideshow.js
0xb0ba: done

oparoz:
Thanks for the fixes.

How long can descriptions be?
I think you might need to use a different CSS style if we're talking about a paragraph.

0xb0ba:
3 seconds as control. enough to read. then goes out to show the full picture.

oparoz:
> 3 seconds as control. enough to read. then goes out to show the full picture.

Wait, so you mean, this is blocking the rendering of the picture? First the description is shown for 3 seconds, then it's replaced by the image and cannot be seen again?

0xb0ba:
no, no. :) description show over pictures, and may close the necessary detail in the picture...
after hide its may be show by mouse movement.

oparoz:
OK, but how big can that description be? If it can be as long as it wants, we may need to limit it to maybe 2 lines and reveal the rest on hover.

0xb0ba:
css already in one line

oparoz:
Yes, but that's for the title, so it's not a problem if we're missing 10% of the filename because there isn't enough space, but if we only get one line out of 10 for a description, then it's going to be frustrating for users.

0xb0ba:
Yes you are right. some people may be stored there stories.

oparoz:
Can you provide a few samples with stories stored in EXIF data?

0xb0ba:
No I can not. I keep a few words describing the photo.

oparoz:
@demattin @setnes @patman15 Anybody with pictures with EXIF descriptions so that we can properly test this PR?

jancborchardt:
@0xb0ba can you post a screenshot of this change so it’s easier to review?

0xb0ba:
default

oparoz:
Thanks for that @0xb0ba. 

Designers need to figure out how to properly show that description. 
- Centred? 
- 3 lines? In this example it would just be the first line over 3 narrower lines
- Scroller or rollover? Rollover would fill the screen with that description since we have to keep it narrow. Scroller requires 2 additional touch areas instead of an ugly scrollbar

A lot of galleries show the exif data in a side-bar. I think it would make it easier to show the meta-data we have.

0xb0ba:
@oparoz, I'm not a designer. I can not answer these questions. in most cases it is the description of a few words, scroller or rollover is likely not needed.

exif data need not always. its cool in popup side-bar. description should always, especially in the slideshow.

jancborchardt:
I’m not sure we should show it there to be honest. What’s the usecase for exif description like that, I have never seen it really. How do other photo apps approach this?

0xb0ba:
> I have never seen it really

for that they need to fill

default

oparoz:
Thanks for the screenshot @0xb0ba. What happens on Picasa when you switch to portrait mode?

0xb0ba:
default

patman15:
@oparoz Sorry, I only have loads of pictures with GPS tags but no comments ...

0xb0ba:
@patman15, gps tags its next step :)

patman15:
@0xb0ba, okidoki let me know when it's ready, I can test that. :-)

oparoz:
Thanks. That works quite well for landscape pictures shown in portrait mode, but would cover 1/3 of a portrait.

0xb0ba:
It's not a problem. it hides after 3 seconds.

oparoz:
The default timer is for the whole interface, so it is a problem when zooming in/out as there is a large block of text masking part of the image.

0xb0ba:
its not problem too - https://github.com/0xb0ba/gallery/commit/add90f0939ea2a62052b6148218b17c11dd8a4ec

0xb0ba:
what is the probability that the block will be large?

DeepDiver1975:

PR #286, opened 2015-08-31
Remove image label gradient in favour of transparent black

tomneedham:
Looks cleaner with the look of OC8.2

Before:
screen shot 2015-08-31 at 16 04 17

After:
screen shot 2015-08-31 at 16 03 53

PR #456, opened 2015-10-27
Adds flickr style gradient to the gallery images.

raghunayyar:
As we discussed at the conference (https://github.com/owncloud/gallery/pull/286), we wanted to use gradients which are something on the lines of Flickr. I have added the exact same gradient here as of now. @jancborchardt would you like to tweak it a bit so that it satisfies all our corner cases?

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -139,10 +139,16 @@ div.crumb.last a {
 
 #gallery .item-container .album-label,
 #gallery .item-container .image-label {
-	background: rgba(0, 0, 0, 0.4);
+	position: absolute;
+	width: 100%;
+	height: 100%;
+	transition: opacity 200ms linear;
+	opacity: 1;
+	background-image: linear-gradient(rgba(0,0,0,0) 65%,rgba(0,0,0,0.35));
 }
 
-#gallery .item-container .image-label .title {
+#gallery .item-container .image-label .title,
+#gallery .item-container .album-label .title {
 	display: inline-block;
 	color: #fff;
 	text-align: center;
@@ -152,9 +158,13 @@ div.crumb.last a {
 	overflow-x: hidden;
 	white-space: nowrap;
 	text-overflow: ellipsis;
+	left: 0;
+	bottom: 10px;
+	position: absolute;
 }
 
-#gallery .item-container .image-label .title:hover {
+#gallery .item-container .image-label .title:hover,
+#gallery .item-container .album-label .title:hover {
 	overflow: visible;
 	white-space: normal;
 }
--- a/js/galleryalbum.js
+++ b/js/galleryalbum.js
@@ -6,10 +6,11 @@
 		'<div class="item-container album-container" ' +
 		'style="width: {{targetWidth}}px; height: {{targetHeight}}px;" ' +
 		'data-width="{{targetWidth}}" data-height="{{targetHeight}}">' +
-		'	<div class="album-loader loading"></div>' +
-		'	<span class="album-label">{{label}}</span>' +
-		'	<a class="album" href="{{targetPath}}"></a>' +
-		'</div>';
+		'<div class="album-loader loading"></div>' +
+		'<a class="album" href="{{targetPath}}">' + 
+		'<div class="album-label">' + 
+		'<span class="title">{{label}}</span>' +
+		'</div></a></div>';
 
 	/**
 	 * Creates a new album object to store information about an album
@@ -62,7 +63,6 @@
 				album.domDef.click(album._showLoader.bind(album));
 
 				album._fillSubAlbum(targetHeight);
-
 				return album.domDef;
 			});
 		},
--- a/js/galleryimage.js
+++ b/js/galleryimage.js
@@ -6,10 +6,11 @@
 		'<div class="item-container image-container" ' +
 		'style="width: {{targetWidth}}px; height: {{targetHeight}}px;" ' +
 		'data-width="{{targetWidth}}" data-height="{{targetHeight}}">' +
-		'	<span class="image-label">' +
-		'		<span class="title">{{label}}</span>' +
-		'	</span>' +
-		'	<a class="image" href="{{targetPath}}" data-path="{{path}}"></a>' +
+		'<a class="image" href="{{targetPath}}" data-path="{{path}}">'+ 
+		'<div class="image-label">' +
+		'<span class="title">{{label}}</span>' +
+		'</div>' + 
+		'</a>' + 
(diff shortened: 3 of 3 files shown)


jancborchardt:
What do you think @oparoz @owncloud/designers?

@raghunayyar can you post a screenshot so more people can review directly?

oparoz:
I haven't looked in details yet. I'm a bit worried about the new width, height and positioning of the element, because it may interfere with future work, so it will take a bit of time to get that checked.

raghunayyar:
Sure : natural state looks like this
screen shot 2015-10-28 at 12 13 16 pm

hovered state looks like this
screen shot 2015-10-28 at 12 13 27 pm

jancborchardt:
@raghunayyar nice! How's it on a brighter image?

css/styles.css
```diff
@@ -130,7 +130,12 @@ div.crumb.last a {
 
 #gallery .item-container .album-label,
 #gallery .item-container .image-label {
-	background: rgba(0, 0, 0, 0.4);
+	position: absolute;
+    width: 100%;
+    height: 100%;
+    transition: opacity 200ms linear;
```
MorrisJobke: Indentation ;)

raghunayyar: I have to figure out things with Atom. lol, will push the fix asap.

raghunayyar:
@jancborchardt usual state
screen shot 2015-10-29 at 9 57 29 pm

hovered state
screen shot 2015-10-29 at 9 57 34 pm

oparoz:
@raghunayyar Would it be possible to see the same images with the current style (pre-PR)? I know it's a bit of work, but it would be great to be able to quickly see the difference when looking at the before and after picture for both the standard and hovered state.

jancborchardt:
Looks awesome! :)

jancborchardt:
Just tested this and it looks quite awesome, way better than the current approach. However, 3 grave problems:
- [x] albums can not be opened, and album titles float on the top. Instead they should be shown on the bottom and have the same fade as for images, but displayed permanently.
- [x] images can not be opened
- [x] the ellipsis of long file names is broken. Should be forced to one line.

oparoz:
Thanks for testing. I was afraid the div changes would break stuff... Hopefully there is a way around that.

jancborchardt:
I guess just changing the opening handler would fix it. But I’m not JS-savvy enough to be sure. ;)

raghunayyar:
@jancborchardt will look into it asap.

oparoz:
There is also the problem of white letters on white background.

whiteonwite

Apart from that, I like the effect, it's more natural than having that blocky background, but it needs to be tuned a bit.

raghunayyar:
@jancborchardt fixed all the issues, do you want to fine tune the gradient? cc @oparoz

jancborchardt:
@oparoz that problem should be fixed if @raghunayyar fixed the ellipsis and forcing the filename display to one line. Then the filename will be shown on the bottom of the fade where it’s enough contrast.

oparoz:
@raghunayyar I apologise, looking at your code, I realised some of my changes had been merged against the wrong branch. Can I ask you to rebase? Since the change is mainly about grouping all the HTML in a template, it should be easy for you to make the small adjustments.

@jancborchardt The problem is not fixed by the title expansion, because the screenshot I've posted is of the extended title. I haven't checked if the gradient expands with the title. If not, then it should. If yes, then the gradient should fade away a bit higher.

jancborchardt:
If there is a longer title, it is either visible in the detail view, or should at most show as a tooltip. It should not result in the name extending to more rows than one in the overview. Ok?

oparoz:
> If there is a longer title, it is either visible in the detail view, or should at most show as a tooltip. It should not result in the name extending to more rows than one in the overview. Ok?

It's not the current behaviour, but could be addressed in a separate PR, yes.

jancborchardt:
Right, but then the issue you mention above should not be taken into account for this PR if we fix it in a different one anyway. ;)

raghunayyar:
@oparoz I have rebased my branch, do you want to tweak the gradient? cc @jancborchardt

css/styles.css
```diff
@@ -152,9 +158,13 @@ div.crumb.last a {
 	overflow-x: hidden;
 	white-space: nowrap;
 	text-overflow: ellipsis;
+	left: 0;
+	bottom: 10px;
+	position: absolute;
 }
 
-#gallery .item-container .image-label .title:hover {
+#gallery .item-container .image-label .title:hover.
```
oparoz: Typo

oparoz:
@raghunayyar - Thanks for the rebase. No need to take care of the gradient since it's been decided to go with a tooltip.
I'm having some issues with the re-ordering of the HTML elements, so I first need to try and fix it before being able to merge this.

jancborchardt:
@raghunayyar also please squash your commits into one.

oparoz:
If someone wants to look at other ways of doing this.
The problem we're having is that when the label covers the whole surface, then it blocks the click handler. Because of old IE support, we can't use pointer-events and need to bubble down the event or place the label inside the anchor. That change is not compatible with the introducton of new features, so the structure should stay the same.

jancborchardt:
Why don’t we have the click handler on the label?

oparoz:
The container would probably be more appropriate, now that we have one.

oparoz:
But it has to be created from scratch since we were simply using anchors

raghunayyar:
Can we merge this since all the discussed stuff has been fixed.

oparoz:
Not with the change of element order as it would need to be reverted in the future.

What we can do is wait for a PR changing the handler to be merged before merging this one.

jancborchardt:
So what’s the plan of action here so we move forward @oparoz @raghunayyar?

oparoz:
1. Add a click handler to the container in this PR or another one which needs to be merged before this one
2. Restore the order of elements in this PR
3. Add a tooltip preferably in this PR or before merging this one so that we don't forget

jancborchardt:
@raghunayyar can you implement what @oparoz described above then? :)

raghunayyar:
@jancborchardt @oparoz I am occupied with calendar at the moment, can you take care, please?

oparoz:
Closing as this has been added to the 9.1-next branch: https://github.com/owncloud/gallery/commit/ee073c99498d932635e3b42197d1496a9434e4bd

oparoz:
Thank you for your contribution!

jancborchardt:
Nice!

PR #457, opened 2015-10-29
Big area for "next" and "prev" buttons. useful for mobile devices.

0xb0ba:


oparoz:
Thanks for your contribution. Have you checked to see if zooming via the scroll-wheel is still working on the desktop? That's the problem we faced with the last PR which was submitted.

0xb0ba:
no, does not work when the cursor is over ".prev" or ".next" or ".menu" or ".name". in this case the prev and next of width 25%

oparoz:
OK, so I've tested this. It's frustrating, because on Chrome, zooming works fine from anywhere, but not on Firefox.

Even though it's only 25% of the horizontal space, I'm still not convinced it won't confuse users. On the one hand, the cursor changes, so it should be obvious that that area can be clicked, but it's a bit annoying to have to keep the cursor in the centre of the image.

@jancborchardt @demattin @setnes Any thoughts for or against this change?

setnes:
I absolutely love this change.  It's a simple way to make things a lot easier on a mobile device.

I agree with @oparoz, about the Chrome vs Firefox zooming though.  This needs to be figured out if at all possible... or maybe balanced with a smaller percentage.

0xb0ba:
OK, let's find a solution for Firefox. I think the problem is "onwheel" and "onmousewheel"

oparoz:
Probably, but just like for pinch-to-zoom which only works on iOS, it may be difficult to fix since the handlers are in Bigshot...

jancborchardt:
Does clicking the image itself also work for advancing the image? That would be really cool as addition. :)

oparoz:
> Does clicking the image itself also work for advancing the image? That would be really cool as addition. :)

That might make it unnecessary to make the buttons larger, but it might not be possible to do due to interferences with Bigshot...

0xb0ba:
> to make the buttons larger

25%+75%

oparoz:
> 25%+75%

That's what use in the other PR, but it blocked the zooming.

setnes:
I tried 15% instead of 25%.  This still gave a much better mobile experience without getting in the way (as much) with zooming with a mouse.  It seems like a better balance given the circumstances.

oparoz:
So @jancborchardt , should we go with 15% even if it blocks scrolling on Firefox or wait for a better solution?

0xb0ba:
@oparoz last commit work in firefox

oparoz:
Ah, I missed that :) Thanks. I'll try asap. That means that we would be able to go to 25-75 is we wanted to.

PVince81:
@oparoz did you get a chance to have a look ?

oparoz:
Thank you for your contribution @0xb0ba. That's really useful.

Merged as-is. A separate PR can be open if the invisible buttons need to be of different dimensions.

oparoz:
This needs to be backported to 8.2.2

jancborchardt:
@0xb0ba @oparoz you can do a backport via git cherry-pick on the commit to stable8.2 :)

Issue #460, opened 2015-11-02
Slideshow will not close if started directly from image link.


setnes:
I didn't notice this when we worked on the back button / slideshow close logic.  This is a subtle annoyance related to linking to images inside a shared link folder.

Here are the steps to reproduce this issue.

Share a directory using a link.  You will have a link that looks like this...
https://files.example.com/index.php/apps/gallery/s/mAg1cT3xT

Now navigate to an image within that folder.  Your link now looks like this...
https://files.example.com/index.php/apps/gallery/s/mAg1cT3xT#example.jpg

If you hit the slideshow X button from here everything works, but if you share this link (or open it in a new browser), the X button will not work.
##

oparoz:
Ah yes, it's probably because there is no history if you share the link directly... Not sure how to fix this.

Maybe I can test to see if the domain is in the history already...

setnes:
I played around with the _exit function in slideshowcontrols.js.  This is very much a brute force hack, so there's no pull request yet.  It reloads the page (bad).  It parses the page URL (bad).  It does what I want (good).

```
 _exit: function () {

  // Only modern browsers can manipulate history
  if (history && history.replaceState) {
    if (window.history.length === 1) {
      // We have nothing to go back to.  Drop the image from the URL and reload the page.
      window.location.assign( window.location.href.slice(0,window.location.href.lastIndexOf(this.images[this.current].name)) );
    } else {
      // We simulate a click on the back button in order to be consistent
      window.history.back();
    }
  } else {
    // For ancient browsers supported in core
    this.stop();
  }

},
```

setnes:
There's a much easier way to do this.  It was right under my nose.  It looks like this.stop() can be called when there is not enough history to use back();  Now it's more of a one line change to the IF condition... and maybe a comment change.  :smile:

```
 _exit: function () {

  // Only modern browsers can manipulate history
  if (history && history.replaceState && window.history.length > 1) {
    // We simulate a click on the back button in order to be consistent
    window.history.back();
  } else {
    // For ancient browsers supported in core
    this.stop();
  }

},
```

oparoz:
This issue was moved to nextcloud/gallery#102

PR #463, opened 2015-11-08
If there is no browser history, exit the slideshow by calling stop.

setnes:
This is a fix for #460.


oparoz:
Thanks, that sounds like a sensible solution :)

Did you test it with various scenarios or just by sharing a Gallery link?

setnes:
I have tried this for awhile, and it seems to be working nicely.  What should be tested specifically?  

I have tried from an image at the top level of a shared link as well as with images in sub-directories of a shared link.  Both work fine.  I have tried from a browser that is logged into an OwnCloud account as well as from browsers that are using only the access that the shared link provides... etc.

oparoz:
> What should be tested specifically?

What you've described + normal use as a logged in user who just browses through his pictures.

PR #470, opened 2015-12-05
No link for last crumb.

### LINKED ISSUE #469: Breadcrumb for current album goes to #.
The href for the current album's breadcrumb is set to "#".  This takes you to the base path when clicked.  The old behavior was to not have the current album's breadcrumb be a link.

setnes:
A little ternary magic fixes #469.

### DIFF
--- a/js/breadcrumb.js
+++ b/js/breadcrumb.js
@@ -141,10 +141,17 @@
 		 * @private
 		 */
 		_pushCrumb: function (name, link, img, cssClass) {
+			var hash = '';
+
+			// Prevent the last crumb from getting a link unless the last crumb is 'home'.
+			if ( cssClass.indexOf('last') === -1 || cssClass.indexOf('home') > -1 ) {
+				hash = '#' + encodeURIComponent(link);
+			}
+
 			this.breadcrumbs.push({
 				name: name,
 				dir: link,
-				link: link ? '#' + encodeURIComponent(link) : "",
+				link: hash,
 				img: img,
 				cssClass: cssClass
 			});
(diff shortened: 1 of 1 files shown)


oparoz:
Thanks!

I'll need to cherry-pick your commit as your branch seems to include all previous commits which haven't been merged yet.

setnes:
Yeah...  I'm dealing with a little rebase nightmare again after trying to cleanup my master.

oparoz:
It's really a pain when working on several large tasks at the same time as they may impact the same files. That's one reason I don't always submit one PR per change when dealing with smaller issues.

setnes:
Yeah.  I'm still new to git.  I'm still trying to undo my mistake where I used my personal master branch for a pull request instead of keeping it identical to upstream.

setnes:
There... that's better.

oparoz:
Well done :)

oparoz:
Doesn't work :( It kills the home icon, so looks like more thoughts will have to be given to that.

setnes:
Hmm... you're right.  We might also have to look at the cssClass to see if it is the home link.

oparoz:
> We might also have to look at the cssClass to see if it is the home link.

Sounds like the easiest solution

setnes:
Try that.  It's not as elegant, but it doesn't destroy the home crumb... even when the home crumb is the last crumb.

oparoz:
I think it won't work because :house: won't have the # needed to load the root folder any more.

setnes:
It does have the #.  Look closer... or try it.  :)

oparoz:
All good thanks :)

setnes:
Cool.  :)

Issue #466, opened 2015-11-30
Clicking on folder label causes spinner to display without actually opening the folder.


setnes:
When I click on folders in my test instance (based on master), I sometimes will never be taken to the folder.  Instead, I will see the new spinner continue to spin.

I haven't figured out what is causing this yet.  Is it somehow related to the new spinner change?

setnes:
I finally figured out how to reproduce this reliably.

If I click on the label (the shaded text area) for an album folder, I will get the spinner indefinitely.  If I click above the album label within the images for the album, the spinner will be displayed, very briefly, before the folder contents are displayed.

So... somehow clicking on the label is different from clicking above the label.  I noticed similar behavior for individual images.  Clicking the text does not open the image, but clicking above the text does.

setnes:
It appears that moving... 

```
<span class="album-label">{{label}}</span>
```

...to be within the hyperlink "a" tag of the TEMPLATE variable defined at the top of galleryalbum.js resolves this problem.  I imagine a similar change might be wanted in galleryimage.js if you also want the image labels to be clickable.

```
        var TEMPLATE =
                '<div class="item-container album-container" ' +
                'style="width: {{targetWidth}}px; height: {{targetHeight}}px;" ' +
                'data-width="{{targetWidth}}" data-height="{{targetHeight}}">' +
                '       <div class="album-loader loading"></div>' +
                '       <a class="album" href="{{targetPath}}">' +
                '               <span class="album-label">{{label}}</span>' + 
                '       </a>' +
                '</div>';

```

@oparoz, let me know what you think of this and I'll see if I can find some time this week to put together a pull request.  I made the mistake of making the last one on my master branch, so I imagine I might need to do some renaming to have two in flight at the same time.  (still git challenged)

oparoz:
Thanks for debugging this issue :)
The way forward is to drop the anchor (a tag) and use a new onClick behaviour for the top container. It should fix both this problem and allow for [better looking labels](https://github.com/owncloud/gallery/pull/456).

You can have a go at that change if you want to, should be straight forward. There are similar examples in the code already.

setnes:
That's a curve ball challenge.  :)

There is already a click tied to _showLoader.  Do you want two click functions... one to call _showLoader and one to call something new like _openAlbum?  I think that would work.

I think we can reference encodeURIComponent(this.path) within an _openAlbum function.  What's the best way to open the page?

[...]

PR #471, opened 2015-12-06
Click handlers for image and album thumbnails.

setnes:
This is the start of #466.  This is not yet complete.  The navigation works using click handlers, but the anchor tags still exist... just without any href.  This allows review of the click logic code without needing to refactor the anchor tag logic and any special CSS references... yet.

Also... using replaceState isn't what we needed here.  We actually do want the URL in the history and a page navigation to occur (not a reload).  A simple location.href assignment makes things work very much like the old links.  The labels are no longer in the way of clicking.

Do we want to test this against the gradient label changes in #456 before refactoring further?

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -139,10 +139,16 @@ div.crumb.last a {
 
 #gallery .item-container .album-label,
 #gallery .item-container .image-label {
-	background: rgba(0, 0, 0, 0.4);
+	position: absolute;
+	width: 100%;
+	height: 100%;
+	transition: opacity 200ms linear;
+	opacity: 1;
+	background-image: linear-gradient(rgba(0,0,0,0) 65%,rgba(0,0,0,0.35));
 }
 
-#gallery .item-container .image-label .title {
+#gallery .item-container .image-label .title,
+#gallery .item-container .album-label .title {
 	display: inline-block;
 	color: #fff;
 	text-align: center;
@@ -152,9 +158,13 @@ div.crumb.last a {
 	overflow-x: hidden;
 	white-space: nowrap;
 	text-overflow: ellipsis;
+	left: 0;
+	bottom: 10px;
+	position: absolute;
 }
 
-#gallery .item-container .image-label .title:hover {
+#gallery .item-container .image-label .title:hover,
+#gallery .item-container .album-label .title:hover {
 	overflow: visible;
 	white-space: normal;
 }
--- a/css/styles.css
+++ b/css/styles.css
@@ -192,7 +192,7 @@ div.crumb.last a {
 	opacity: .5;
 }
 
-#gallery a.image > img {
+#gallery .image-container .image > img {
 	max-height: 200px;
 }
 
--- a/js/galleryalbum.js
+++ b/js/galleryalbum.js
@@ -6,9 +6,12 @@
 		'<div class="item-container album-container" ' +
 		'style="width: {{targetWidth}}px; height: {{targetHeight}}px;" ' +
 		'data-width="{{targetWidth}}" data-height="{{targetHeight}}">' +
+		'	<a class="album" href="{{targetPath}}">' +
+		'		<div class="album-label">' +
+		'			<span class="title">{{label}}</span>' +
+		'		</div>' +
+		'	</a>' +
 		'	<div class="album-loader loading"></div>' +
-		'	<span class="album-label">{{label}}</span>' +
-		'	<a class="album" href="{{targetPath}}"></a>' +
 		'</div>';
 
 	/**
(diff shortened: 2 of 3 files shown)


oparoz:
Thanks!

Testing #456 would be great. It should simply be a matter of merging the CSS changes.

setnes:
I merged the changes from #456 into this.  It seems to work.

I think the album label needs to be somewhat larger... or at least somehow different than the image label.  A folder with just one picture is a good test case for this.  Also, should there be CSS to change the cursor to "pointer" when hovering over an image or album?

The anchor tags still need to be refactored.  Right now they are still just placeholders without any href defined so we can prove it all works.

oparoz:
Looking at the commits, it seems you took everything in #456, but we only need the CSS. It's important (for the test) because the element swapping is what is messing future plans up.

> I think the album label needs to be somewhat larger.

I did try to use something different, but couldn't find a satisfactory solution, mostly because we can't use bold and because albums and pictures come in many different sizes.

> Also, should there be CSS to change the cursor to "pointer" when hovering over an image or album?

It's like that on master, so it has to be the same in this PR. I haven't checked where it's missing.

setnes:
We need more than the CSS to make the change work.  For example, the CSS is referencing the title class under the album-label class.  I needed to merge some of the changes into my galleryalbum.js and galleryimage.js or it would not work.  No worries.

Adding a "cursor: pointer;" to the CSS is all that is needed to fix the cursor.  The cursor changes in master only because of the anchor tags.

setnes:
I have refactored the anchor tags.  Everything seems to be working with the exception of keyboard navigation.  Give it a try.

oparoz:
## Acceptance tests
- [ ] Clicking on album opens the album
- [ ] Clicking on album label opens the album
- [ ] Clicking on an album shows a spinner
- [ ] Clicking on image launches the slideshow
- [ ] Clicking on image label launches the slideshow
- [ ] If an album contains a single bad JPEG, the media type icon is shown
- [ ] Clicking on an album containing a single bad JPEG opens the album. For testing purposes, please only put a couple of pictures in the album
- [ ] Clicking on that album shows a spinner
- [ ] If an album contains a bad JPEG, that image is removed from the representation
- [ ] Clicking on an album containing a bad JPEG opens the album, as usual
- [ ] Clicking on that album shows a spinner
- [ ] Future plans aren't ruined (Internal)

setnes:
Should we really be doing this?  We can make it work with an anchor tag.

This is pretty insightful.  http://stackoverflow.com/a/32194060

oparoz:
Thanks for digging that out. I only looked at performance comparisons when trying to decide if switching was wise, but this seems like a valid point, especially for accessibility reasons.
So maybe we should wrap each div in an anchor.

setnes:
Yeah.  I think we can have both the anchor tag and a click handler.  I want a click handler on the image anyway so I can set a variable to fix the issue with closing the slideshow when there is no history.  We already have a need for the click handler on the album to show the spinner.

oparoz:
Cool. I hope it won't be a headache with getting the size of elements and resizing them. We'll see :)

setnes:
The latest commit is working with the anchor tags and label gradients.
I lost evidence that @raghunayyar made the CSS changes when I rebased my branch.  I think this all needs to be merged together though.  Should I try to fix that?

oparoz:
In your commit, you wrapped the label with the anchor, but we need to wrap the container or we're back to square one.

Something like this in galleryimage.js:

```
    <a href="{{targetPath}}">
        '<div class="item-container image-container" ' +
        'style="width: {{targetWidth}}px; height: {{targetHeight}}px;" ' +
        'data-width="{{targetWidth}}" data-height="{{targetHeight}}">' +
        '   <span class="image-label">' +
        '       <span class="title">{{label}}</span>' +
        '   </span>' +
        '   <span class="image" data-path="{{path}}"></span>' +
        '</div>' +
    '</a>';
```

Hopefully this doesn't introduce any new issues.

setnes:
I did mess with that, but I was having problems having the album-loading-spinner-thingy show up correctly if it was nested inside the anchor tag.  The code in my branch is working very well for me.  Do you have future plans where getting rid of the parent div tag is important?

setnes:
I've gone through all of the tests except for "future plans" one.  :)

oparoz:
Thanks!

The one plan where there is a conflict is the "instant thumbnails" PR. With your changes, the labels are resized with the images and we don't want that.

setnes:
I made the anchor the outside element.  It now has the container classes as the div was suddenly redundant.  I hope this helps.

This is passing the same tests as before.  Also, we should think about removing the 1px margin between the cropped images in the album link.  The label shows better that way.  (not in this change)

_Make up your own snarky comment about how this pull came first._  :smirk:

oparoz:
Closing as this has been added to the 9.1-next branch: https://github.com/owncloud/gallery/commit/a3a3a5051b77b9233ca816f51e01200df421ef0a
Thank you for your contribution!

PR #477, opened 2015-12-13
More param validation

### LINKED ISSUE #445: gallery.cnf is not verifying "sort" button
Data such as "type" is not checked properly, due to some hardening in core this does not lead directly to an exploitable XSS but should get addressed anyways.

Related code: https://github.com/owncloud/gallery/blob/d904a6a4af293e5730e397218c7e0713c3461215/js/galleryview.js#L322

Probably we should verify in the controller and also in the JS for data from the gallery config.

oparoz:
It's a good idea to validate user fed parameters before sending them to the browser...

Fixes #445

@LukasReschke

### DIFF
--- a/config/configparser.php
+++ b/config/configparser.php
@@ -142,10 +142,9 @@ class ConfigParser {
 	 */
 	private function buildAlbumConfig($currentConfig, $parsedConfig, $completionStatus, $level) {
 		foreach ($completionStatus as $key => $complete) {
-			if (!$this->isConfigItemComplete($key, $parsedConfig, $complete)
-			) {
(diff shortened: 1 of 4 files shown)


oparoz:
- [ ] Optimise and clean-up (scrutinizer + squash)

oparoz:
@LukasReschke - This needs to be reviewed soon if it's to make the 8.2.3 release

LukasReschke:
Good stuff. I'd however recommend not to backport it since there can't be done too many shenanigans with it and that's a rather huge change :smile: 

:+1:

oparoz:
Thanks for the review! :)

PR #481, opened 2015-12-24
Improve SVG cleanup after sanitization

### LINKED ISSUE #467: Update DOMPurify and use @oparoz's SVG hook
This new version will break less SVGs while still properly sanitizing them.

oparoz:
Fixes #467 

I've loaded my standard samples as well as an "armed" SVG and there is only one SVG which can't be properly rendered. It seems impossible to get it to display properly without opening the gate as it contains internal references.


oparoz:
@LukasReschke if you could test this to make sure it's still as safe as before. The people at DOMPurify seem to think so.

oparoz:
Ping @LukasReschke

oparoz:
In short:
- The custom hooks have been moved to the utility class, to be shared between the slideshow and the gallery app
- Illustrator doctype is replaced with standard one
- Filter tags are allowed in SVGs 
- Garbage characters left at the start of a purified SVG are removed
- DOM purify has been updated to 0.7.3

LukasReschke:
Great stuff! :+1:

oparoz:
Thanks for the review! :)

PR #485, opened 2015-12-28
Add uploading (drag and drop) support for registered users

oparoz:
Superseded by https://github.com/owncloud/gallery/pull/520

### DIFF
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -82,7 +82,10 @@ class PageController extends Controller {
 		$appName = $this->appName;
 
 		// Parameters sent to the template
-		$params = ['appName' => $appName];
+		$params = [
+			'appName' => $appName,
+			'uploadUrl' => $this->urlGenerator->linkTo('files', 'ajax/upload.php')
+		];
 
 		// Will render the page using the template found in templates/index.php
 		$response = new TemplateResponse($appName, 'index', $params, 'public');
--- a/css/public.css
+++ b/css/public.css
@@ -57,10 +57,3 @@ body {
 #gallery.hascontrols {
 	padding-bottom: 0;
 }
-
-#loading-indicator {
-	height: 32px;
-	width: 100%;
-	margin-top: 100px;
-	margin-bottom: 0;
-}
--- a/css/styles.css
+++ b/css/styles.css
@@ -152,10 +152,16 @@ div.crumb.last a {
 
 #gallery .row > a > .album-label,
 #gallery .row > a > .image-label {
-	background: rgba(0, 0, 0, 0.4);
+	position: absolute;
+	width: 100%;
+	height: 100%;
+	transition: opacity 200ms linear;
+	opacity: 1;
+	background-image: linear-gradient(rgba(0,0,0,0) 65%,rgba(0,0,0,0.35));
 }
 
-#gallery .row > a > .image-label .title {
+#gallery .row > a > .image-label .title,
+#gallery .row > a > .album-label .title {
 	display: inline-block;
 	color: #fff;
 	text-align: center;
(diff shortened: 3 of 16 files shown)


jospoortvliet:
I'd suggest to split this, then, start with the drag-and-drop and do the + button later. That way, we not only make the whole thing a bit smaller (easier to review and contribute to) but also ppl can start testing the drag-and-drop part. And, of course, if you don't have time to finish the 2nd part, the first is in.

jancborchardt:
Yeah, good call. Even though I’d say the implementation via + button is more important and drag&drop is a nice addition for more advanced folks.

oparoz:
> I'd suggest to split this

No problem from my side. Also, this is a pre-requisite for drag and drop because it touches the way we collect information, so the quicker this is in, the quicker drag and drop can be implemented.

> Even though I’d say the implementation via + button is more important and drag&drop is a nice addition for more advanced folks

I think it's more about discovery, as you've mentioned in the issue itself. Some people won't know it's there unless there is a button or a message telling them things have changed.
Having said that, I expect new users to intuitively try drag and drop and to be pleasantly surprised and older users would know about it because of @jospoortvliet 's special report on the new features ;)

In any case, the important first step is to (in)validate the approach.

oparoz:
OK, so is it working as expected at least? :P

jospoortvliet:
I tried to apply the patch to my stock ownCloud 8.2.2 Gallery app, got this:

```patch -p1 -R < 14609ac5444d6f455220ceaed518ef4081795d42.patch
patching file controller/pagecontroller.php
patching file css/styles.css
Hunk #1 succeeded at 288 (offset 16 lines).
patching file js/gallery.js
patching file js/galleryimage.js
Hunk #1 succeeded at 4 (offset -11 lines).
patching file js/galleryview.js
Hunk #1 FAILED at 9.
Hunk #2 succeeded at 81 (offset -2 lines).
Hunk #3 succeeded at 210 (offset -2 lines).
1 out of 3 hunks FAILED -- saving rejects to file js/galleryview.js.rej
patching file js/thumbnail.js
patching file js/upload-helper.js
patching file service/searchmediaservice.php
patching file templates/part.content.php
Hunk #1 FAILED at 27.
Hunk #2 succeeded at 43 (offset -9 lines).
Hunk #3 succeeded at 69 (offset -9 lines).
Hunk #4 succeeded at 103 (offset -9 lines).
1 out of 4 hunks FAILED -- saving rejects to file templates/part.content.php.rej
patching file tests/unit/GalleryUnitTest.php
patching file tests/unit/controller/PageControllerTest.php
patching file tests/unit/service/SearchMediaServiceTest.php

```

I'm going to guess that with some fails this won't work well...
```

oparoz:
> I'm going to guess that with some fails this won't work well...

Correct. You need to use the 9.0 VM

oparoz:
@setnes @demattin @arkascha @elpraga @sualko @bugsbane @mmattel - This needs testing :)

Bugsbane:
@oparoz Sure. Can you just tell me how to apply the correct patch?

oparoz:
Look in the block below, click on the _command line instructions_ and follow step 1.

oparoz:
That's if you're using a git test instance, of course. I'm not sure it would work as-is on 8.2

Bugsbane:
I don't actually see a "block below", and I am on 8.2.2 rather than a git version. I'm fine with backing up the database and app files. I might be able to set up a second (git) install though.

oparoz:
Here is [the official help with detailed instructions](https://help.github.com/articles/checking-out-pull-requests-locally/). Only step 1 is required. Click on the tiny screenshots to see what we're talking about.

If you don't want to setup a 2nd instance you can use [daily VMs](https://en0ch.se/index.php/s/QB2mI2GsrN5FVtX). They're great for testing as they're pretty much plug and play.
1. Mount the .vmdk in VirtualBox
2. Login and become root in Ubuntu Server (user: ocadmin, pwd: owncloud)
3. ctrl+c to abort the script
4. `$ bash`
5. `cd /var/www/`

PR #491, opened 2016-01-08
Show slideshow images in fullscreen

RoxasShadow:
The original image is 1200x849, but the slideshow applies CSS rules to show it smaller. I know that I can zoom in, but it's annoying since the resize doesn't try to fill the viewport as much as possible.

_First_

``` html
<img src="/index.php/apps/galleryplus/preview/59594?c=fc01f2275f4c1ebdc962202b35307850&amp;width=2900&amp;height=2900&amp;requesttoken=..." alt="51284906_p0_master1200.jpg" style="position: absolute; background-color: rgb(0, 0, 0); background-size: 660px 466.95px; top: 46px; left: 462px; background-position: 0px 0px; height: 466.95px; width: 660px;">
```

screen shot 2016-01-07 at 11 49 25

_Then_

``` html
<img src="/index.php/apps/gallery/preview/59594?c=fc01f2275f4c1ebdc962202b35307850&amp;width=2900&amp;height=2900&amp;requesttoken=..." alt="51284906_p0_master1200.jpg" style="position: absolute; background-color: rgb(0, 0, 0); background-size: 766.078px 542px; top: 0.28033px; height: 542px; left: 409.28px; width: 766.078px; background-position: 0px 0px;">
```

screen shot 2016-01-08 at 01 22 23

Latest stable versions of OSX and Firefox.

### DIFF
--- a/js/slideshowzoomablepreview.js
+++ b/js/slideshowzoomablepreview.js
@@ -50,8 +50,8 @@
 				this.zoomable = null;
 			}
 			var maxZoom = this.maxZoom;
-			var imgWidth = image.naturalWidth / window.devicePixelRatio;
-			var imgHeight = image.naturalHeight / window.devicePixelRatio;
+			var imgWidth = image.naturalWidth;
+			var imgHeight = image.naturalHeight;
 			// Disable zooming in IE when we can't get the image's size (SVG)
 			if (imgWidth === 0) {
 				$(image).attr('width', '100%')


oparoz:
The problem with automatically blowing up pictures is that it makes them blurry and doesn't let the user know what the real size is.
Per example, if you're watching a smaller version of the original, you would only know because it's blurry, not because of its size.

We need more feedback on this behaviour, but it won't be solved this way, because we need the ratio for mobile.
I'm leaving this open so that @jancborchardt and friends can provide their valuable input.

jancborchardt:
Smaller images should not be stretched to fill the view by default. The default is simple: show them at 1:1 size, until it hits either width or height of viewport. Anything beyod can be done via zoom.

RoxasShadow:
@jancborchardt this is exactly what I want and I'm doing right now, but it's not the default behavior (or, at least, it doesn't always work in the correct way) as you can notice by the picture I attached to the OP.

oparoz:
The current behaviour is correct.

Your image is 1200x849, which means that on your 4k monitor you could fit it about 3 times along the horizontal axis. Looking at your first screenshot, we can see that it's the case.

When you fill the screen with the image, you're actually degrading the quality of the image. It might be more obvious with some other content.

RoxasShadow:
@oparoz I'm using a MacBook Pro with resolution set to 1280x800. I just expect to see the image filling as much space as possible, keeping of course the aspect ratio.

oparoz:
@RoxasShadow That's not what the Javascript is detecting. What do you get on this site: ~~removed~~ ?

oparoz:
Use this one:
http://www.mydevice.io/

What do you have for size and pixel ratio?

oparoz:
Just thought of something... Maybe you have a monitor connected...

oparoz:
> MacBook Pro (Retina, Mid 2012) and MacBook Pro (Retina, 15-inch, Early 2013) and later displays have a 2880-by-1800 native resolution at 220 pixels per inch with support for millions of colors.

https://support.apple.com/en-us/HT202471

Ah, but the 13" non-retina is 1280 by 800 (native)

RoxasShadow:
@oparoz yes, I have the 15" one but I [downscaled](https://cloud.githubusercontent.com/assets/805144/12271093/c59430d0-b959-11e5-9d39-7027332472af.png) the resolution.

oparoz:
Moving this to https://github.com/owncloud/gallery/issues/495 since this PR won't fix the problem.

PR #497, opened 2016-01-14
Instant thumbnails showing

oparoz:
See #511

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -50,10 +50,3 @@ body {
 #gallery.hascontrols {
 	padding-bottom: 0;
 }
-
-#loading-indicator {
-	height: 32px;
-	width: 100%;
-	margin-top: 100px;
-	margin-bottom: 0;
-}
--- a/css/styles.css
+++ b/css/styles.css
@@ -278,12 +278,6 @@ http://www.bypeople.com/author/comatosed/*/
 	border-color: #bbb #bbb transparent #bbb;
 }
 
-#loading-indicator {
-	height: 32px;
-	width: 100%;
-	margin-bottom: 50px;
-}
-
 .icon-gallery {
 	background-image: url('../img/gallery-dark.svg');
 }
--- a/css/styles.css
+++ b/css/styles.css
@@ -182,13 +182,18 @@ div.crumb.last a {
 	opacity: .5;
 }
 
+#gallery a {
+	-webkit-transition: all .2s ease-in-out;
+	transition: all .2s ease-in-out;
+}
+
 #gallery a.image > img {
 	max-height: 200px;
 }
 
 #gallery .cropped,
 #gallery .row {
-	opacity: 1;
+	opacity: 0.5;
 	-webkit-transition: opacity 500ms;
 	transition: opacity 500ms;
 }
@@ -199,6 +204,16 @@ div.crumb.last a {
 	margin: 1px;
 	background-position: center;
 	background-size: cover;
+}
+
+#gallery a.album .icon-loading,
+#gallery .image-container .icon-loading {
+	margin: auto;
+	position: absolute;
+	top: 0;
+	left: 0;
+	bottom: 0;
+	right: 0;
 }
 
 #gallery > h2 {
--- a/css/styles.css
+++ b/css/styles.css
@@ -183,8 +183,14 @@ div.crumb.last a {
 }
 
 #gallery a {
-	-webkit-transition: all .2s ease-in-out;
-	transition: all .2s ease-in-out;
+	-webkit-transition: transform .2s ease-in-out, filter 1s ease-in-out, -webkit-filter 1s ease-in-out, -moz-filter 1s ease-in-out, -ms-filter 1s ease-in-out, -o-filter 1s ease-in-out;
+	transition: transform .2s ease-in-out, filter 1s ease-in-out, -webkit-filter 1s ease-in-out, -moz-filter 1s ease-in-out, -ms-filter 1s ease-in-out, -o-filter 1s ease-in-out;
+	
+	filter: gray;
+        -webkit-filter: grayscale(100%) brightness(200%);
+        -moz-filter:    grayscale(100%);
+        -ms-filter:     grayscale(100%);
+        -o-filter:      grayscale(100%);
 }
 
 #gallery a.image > img {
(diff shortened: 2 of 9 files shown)


oparoz:
@jospoortvliet @setnes @demattin you can play with it and tell me what you think.

@jancborchardt Let me know if this is acceptable for Gallery.

jospoortvliet:
The way you describe it makes me want to say "yes!!!" and I could give it a try if it applies to gallery for 8.2 ;-)

I don't have 9.0 anywhere atm...

but perhaps you can show a screenshot or two and we can all just say :+1: on the design ;-)

code-wise, not much I can say anyway.

oparoz:
@jospoortvliet Unfortunately, it's all about the visual special effects, so it only works if you see it for yourself and it's important to use it with your hardware to be able to tell if there are issues...

Just grab the 8.2.2 vm and clone master. That should do it :)

setnes:
When I load this after deleting the cached thumbnails, I get random broken image icons (just some of the images).  When I reload the page everything works.

On mobile it breaks badly once there are enough images to necessitate scrolling.  The images are covered up by the page footer and there's nothing left to see after it reaches that point.

I didn't particularly like the spinner.  So I like that you're looking at changing this.  I find this a bit jittery though.  It's cool once all the thumbnails are cached, but on first load it seems to cause a lot of unnecessary movement and overlapping while things slowly get created.  It makes me feel like something is broken... then it magically snaps into place.

setnes:
The mobile issue might be a core issue with 8.2.2.  It happens in the file view as well as the gallery view.  It affects shared link views.  It does not happen if I am logged in.  

I often test mobile with a shared link to avoid typing a ridiculous password.

oparoz:
> On mobile it breaks badly once there are enough images to necessitate scrolling. The images are covered up by the page footer and there's nothing left to see after it reaches that point.

Are you sure it's not because you have the pdfviewer app installed? It breaks other apps.

> on first load it seems to cause a lot of unnecessary movement and overlapping while things slowly get created. It makes me feel like something is broken... 

That's my main issue with this. On desktop, it's OK as there is enough power to both show and move images, but on mobile, there is a little bit of stuttering... I probably need to throttle the rendering or make sure only one CSS effect is applied at a time.

setnes:
Shutting off the PDF Viewer app solved both the footer problem and the broken image problem.  I never would have guessed that one.  Is this a known issue?

I actually find the image overlapping more awkward on desktop than on mobile.  The browser width on my phone is narrow enough that there's usually only a couple of images in each row.  On a larger screen there are more images per row and the effect is somewhat less attractive.  Don't get me wrong... I think this could be made better than the spinner.

setnes:
There is an eventual fix for the PDF Viewer bug.  See owncloud/core#21373, and owncloud/files_pdfviewer#91.

oparoz:
>  Is this a known issue?

As you've figured out, it is :)

> I actually find the image overlapping more awkward on desktop

Well, you can see the code. If you have any idea on how to improve this let me know.
One possible solution is to not show any images before the resizing and then reveal them, but the spinners will still be going through the same motion as the point is to show that something is happening, so I'm not sure it would help that much.

When looking for improvements, the thing you have to bear in mind is that we don't have the size of the images, so the scaffolding is going to be wider than the page and it can't be resized until we have a full row.

setnes:
It turns out the PDF Viewer issue really has nothing to do with the broken images.  There has to be a race condition in the code somewhere.  That's probably more pressing than how the overlapping images look.  It happens when I delete the thumbnails and reload a page with many images.

I did have an odd thought about using the images as the basis of a progress bar.  The images would go back to not being displayed until the row is complete.  With some masking, however, it should be possible to use their positions (temporary as they are) to create some sort of special effect that looks like a progress bar.  That's probably not where you were going with this though as it would be a lot of work to get something very similar to the spinner.  :)

Here's an example of the broken images I see.  I'm going to switch back to master and see if I can reproduce the problem there before assuming the problem is from this change.

broken_images

setnes:
I don't know if it's good news or bad news, but I was able to reproduce the broken image problem without this change.

oparoz:
Regarding the broken images. If you did a full refresh, it would be interesting to get the errors from the browser and owncloud.log (in a separate issue). There can be many reasons for this.

And well, a progress bar is not a bad idea, it's more useful than a spinner, that's for sure, but there is a similar issue in that we don't know how many images will fit, so we don't know where 100% is.

setnes:
I figured out that setting filesystem_check_changes to 1 fixes the problem.  I guess that should make some sense as I'm deleting thumbnails directly from the data directory on my test instance in order to try out your change.  When filesystem_check_changes is set to 0 (the default), the ownCloud system assumes they should exist.  Then, when they're requested, there's a race condition while they get created by the preview stuff.  Watching the log helped debug that.  There's no bug here.  :)

oparoz:
Ah, I see. You should use `$ sudo -uwww-data ./occ gallery:delete-thumbnails myuser` ;)

setnes:
Yeah.  I don't have that yet... but it's exactly what I should use in the future.  :)

Spacefish:
I doubt this change has anything to do with broken thumbnails, as it does not change any php code.. The broken thumbnails are failed request which just return empty pages / 503 pages, so i guess it´s a php script problem.
It feels much more fluid with the spinners and blend in animations.. I vote for merging this!

I dislike the resize animation after load.. it would be better if the images are loaded in the correct size from the start

oparoz:
@Spacefish - Is your commit an attempt to fix this?

> I dislike the resize animation after load.. it would be better if the images are loaded in the correct size from the start

Or are you talking about something else?

Regarding mobile. I'm thinking of maybe throttling the appearance of rows to make sure the browser doesn't try to do too much at once.

oparoz:
BTW, I'm not getting any grayscale images with this patch

oparoz:
Rows are now throttled. Let me know what you think.

Regarding the grayscale, it doesn't work on Firefox. Also, although I like the effect, I'm concerned about the extra CPU+GPU requirements. The goal of this PR is to remove what's blocking the rendering of images. I think we can make it fancier later, no?

oparoz:
@demattin @arkascha @elpraga @sualko @bugsbane @mmattel - This needs testing :)

Bugsbane:
@oparoz Sure. Can you just tell me how to apply the correct patch?

oparoz:
Superseded by #511

Issue #500, opened 2016-01-23
Try to fix the horrible scrolling performance on mobile introduced by core on 8.2


oparoz:
I don't remember all the details, but I think that from 8.2 the controls are in a fixed div which kills all hardware acceleration on mobile.
- [x] Find out why this was introduced exactly
- [x] Fix it if we don't have a reason to use the same CSS
### Desktop
- Chromium/win10: no visible change
- Firefox/win10: no visible change
- Edge/win10: no visible change
- IE11/win10: no visible change
### Mobile
- Chrome/ios9: huge performance gains. Activates smooth kinetic scrolling
- Safari/ios9: huge performance gains. Activates smooth kinetic scrolling with bouncing edges
- Browser/BB10: improves scrolling performance
- Chrome/Android 4.x: huge performance gains. Activates smooth kinetic scrolling
- Firefox/Android 4.x: no visible change. Seems like hardware acceleration is not available on 4.x
- Chrome/Android 5.x
- Firefox/Android 5.x
- Chrome/Android 6.x
- Firefox/Android 6.x

oparoz:
I think there is hope :). It involves reorganising the templates. Hopefully this won't lead to to much code re-writing.

oparoz:
Re-opening this to give a chance to people to voice their opinion if something doesn't seem right on Android 5+

davivel:
I'll have a look.

ghost:
@davivel any update?

davivel:
Sorry, @cmonteroluque , I forgot this one :$. I'll have look today.

davivel:
@oparoz , I'd need some more details about how did you compare peformance. 

I tried this:
- Server 1: OC server master + gallery app v9.0 installed via git, folder with over 200 pictures from Nexus camera
- Server 2: OC server 8.2 + gallery app v8.2 installed via git, folder with same pictures from Nexus camera
- client: Nexus 5 with Android 6, Chrome 49.0.2623.105 ; I use two devices with same configuration to access at the same time to server 1 and server 2 in each of them and compare visually.
- Steps: Access to folder in Chrome, enable 'grid mode', scroll up & down several times.

I don't find performance problems with none of them, what makes me suspect that my environment or my test is not appropriate. Please, tell me what you think about it.

[...]

PR #507, opened 2016-02-09
Bring back smooth scrolling

oparoz:
First attempt at fixing #500.

The improvements range from _not visible_ to _astonishing_ on the devices I've tested. The devices which benefit the most seem to be iOS based.

This introduces the changes that @fuhaoshih tried to introduce with https://github.com/owncloud/gallery/pull/285, but without the side effects (at least in the scrolling).

**The best quick test you can do is scroll through a list of files on the Files side and then switch to Gallery and scroll again.**

The standard and public templates should be combined in a future PR as they're very similar now.
### Desktop
- Chromium/win10: no visible change
- Firefox/win10: no visible change
- Edge/win10: no visible change
- IE11/win10: no visible change
### Mobile
- Chrome/ios9: huge performance gains. Activates smooth kinetic scrolling
- Safari/ios9: huge performance gains. Activates smooth kinetic scrolling with bouncing edges
- Browser/BB10: improves scrolling performance
- Chrome/Android 4.x: huge performance gains. Activates smooth kinetic scrolling
- Firefox/Android 4.x: no visible change. Seems like hardware acceleration is not available on 4.x
- Chrome/Android 5.x
- Firefox/Android 5.x
- Chrome/Android 6.x
- Firefox/Android 6.x

---

@jancborchardt @henni @setnes @demattin @mmattel @Spacefish @Bugsbane @MorrisJobke @jospoortvliet

### DIFF
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -85,7 +85,7 @@ class PageController extends Controller {
 		$params = ['appName' => $appName];
 
 		// Will render the page using the template found in templates/index.php
-		$response = new TemplateResponse($appName, 'index', $params);
+		$response = new TemplateResponse($appName, 'index', $params, 'public');
 		$this->addContentSecurityToResponse($response);
 
 		return $response;
--- a/css/public.css
+++ b/css/public.css
@@ -4,6 +4,15 @@ body {
 	width: 100%;
 }
 
+#content-wrapper {
+	background-color: #fff;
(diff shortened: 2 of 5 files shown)


oparoz:
I've performed some tests to make sure it wouldn't block any future template extension and we will have to write our own CSS in order to be able to keep a smooth photowall or wait for `core` to fix their templates before trying to add a navigation panel.

PR #511, opened 2016-02-10
Instant thumbnails showing

### LINKED ISSUE #29: Show thumbnails as soon as they're received

oparoz:
Replaces #497

Improves the feedback shown to users to let them know that something is happening when thumbnails have not been generated yet. This replaces the spinner added below the row.

Fixes #29

The improvements are better seen on albums which contain sub-albums as these should show some activity before the full row has been rendered.
### Screencasts

_Note: the refresh rate in the videos is quite low, so this is not reflective of the smootheness of the animation_

**With cached thumbnails**
http://screencast-o-matic.com/watch/cDn6n7hR39

You can see that once the album is cached, there is no resizing taking place any more

**On an empty cache**
http://screencast-o-matic.com/watch/cDn6nEhR3E
### Features:
- Semi-opaque row while it's being filled
- Spinners on images in albums
- Spinners on single images on row (with caveat, see below)
### Caveats
- Since we don't have the real size of the images, the next image or its spinner are sometimes misaligned. It's currently impossible to fix and has to be accepted as a side-effect of being able to show some information before having the full picture
- When resizing the window on a desktop, there is a flashing effect as each time the wall is being redrawn, the opacity goes back to being semi-transparent until the images are resized. In my experience, the fix involves introducing and monitoring a state to know if we're currently in a resizing loop. I don't think it's worth for me to invest time to improve this, but a PR can certainly be introduced later if someone finds a better way
- On low powered devices, there may be a bit of stuttering when resizing very wide images. I find this acceptable, compared to the advantages.
### Tested on
- Windows 10 Firefox, Chrome
- OSX Chrome
- BlackBerry 10
- iOS
- Android 4.3

---

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -57,10 +57,3 @@ body {
 #gallery.hascontrols {
 	padding-bottom: 0;
 }
-
-#loading-indicator {
-	height: 32px;
-	width: 100%;
-	margin-top: 100px;
-	margin-bottom: 0;
-}
--- a/css/styles.css
+++ b/css/styles.css
@@ -285,12 +285,6 @@ http://www.bypeople.com/author/comatosed/*/
 	border-color: #bbb #bbb transparent #bbb;
 }
 
-#loading-indicator {
-	height: 32px;
-	width: 100%;
-	margin-bottom: 50px;
-}
-
 .icon-gallery {
 	background-image: url('../img/gallery-dark.svg');
 }
--- a/css/styles.css
+++ b/css/styles.css
@@ -189,13 +189,18 @@ div.crumb.last a {
 	opacity: .5;
 }
 
+#gallery a {
+	-webkit-transition: all .2s ease-in-out;
+	transition: all .2s ease-in-out;
+}
+
 #gallery a.image > img {
 	max-height: 200px;
 }
 
 #gallery .cropped,
 #gallery .row {
-	opacity: 1;
+	opacity: 0.5;
 	-webkit-transition: opacity 500ms;
 	transition: opacity 500ms;
 }
@@ -206,6 +211,16 @@ div.crumb.last a {
 	margin: 1px;
 	background-position: center;
 	background-size: cover;
+}
+
+#gallery a.album .icon-loading,
+#gallery .image-container .icon-loading {
+	margin: auto;
+	position: absolute;
+	top: 0;
+	left: 0;
+	bottom: 0;
+	right: 0;
 }
 
 #gallery > h2 {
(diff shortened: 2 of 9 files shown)


oparoz:
@Spacefish - I removed your grayscale filter as it wasn't working on Firefox and in my tests on mobile, making things slower, but this can be re-added in a separate PR where we can better test it in isolation.

sualko:
Seems to work in Chrome 48.0.2564.103 (64-bit) on Mac OS with oc 9.0-pre.

oparoz:
This is now part of the [next-9.1 branch](https://github.com/owncloud/gallery/commits/next-9.1), but I'm leaving it open so that people can still express themselves.

PR #514, opened 2016-02-11
Remove IE8 support in the slideshow

oparoz:
Fixes https://github.com/owncloud/gallery/issues/516


oparoz:
Normal IE still works. IE8 doesn't work at all on Files, so this is safe to merge.

PR #519, opened 2016-02-12
Add drag and drop support

oparoz:
Replaced by #581


jancborchardt:
Nice! :) Only some details:
- the dashed border around albums looks a bit off. What about dotted, and rather a color of rgba(0,0,0,.5)?
- the message »target exists« is too technical. Should be »File already exists.« (or Folder if it’s a folder). Or could we have a replacement dialog like in Files? cc @butonic

oparoz:
> the dashed border around albums looks a bit off. 

I agree. 

> What about dotted, and rather a color of rgba(0,0,0,.5)?

Dotted didn't work in my previous tests, but I'll try with your colour suggestion.
Maybe There should just be a plain colour, but that might look strange with transparent pictures

> the message »target exists« is too technical. Should be »File already exists.« (or Folder if it’s a folder). Or could we have a replacement dialog like in Files? 

I've copied the behaviour and message from Files, so it's best to open an issue there and then I can sync with any changes made there. It's not possible to have a pre-move conflict dialogue as the content of the target is not known beforehand. An improvement in Files could be to have a post-move dialogue.

oparoz:
Tried dotted again and saw why it doesn't work. Each side is dotted instead of having one dotted line going all the way around the square. This makes the corners look ugly. By using a dashed line, the problem is almost non-existent.

oparoz:
I've improved the look of the drop target. I think it looks good enough now.
Bonus: mobile support.

oparoz:
On mobile, you need to tap and hold to be able to move elements. It has to be done this way in order to not have a conflict with scrolling.

oparoz:
This could also use some testing: @tahaalibra @viraj96 @imjalpreet @mbtamuli

tahaalibra:
I tested and found various issues:

## Public
- the thumbnail view is gone and plain list is started as default
- when i changed the view to thumbnail...i can drag and drop the photo..although it does not succeed, i think we should remove drag and drop from public view

## Logged In
- when i try moving a folder f1 to another folder f2. i get a js error
  "gallery.js?v=22125a1…:394 Uncaught TypeError: targetPath.charAt is not a function"
  also the loading animation on f2 does not go away(i thinks this is because of the js error)
  selection_002

NOTE: this is happening only when i try to move one album into another at the fist page of gallery(outside of all albums)

-when i move photos using breadcrumbs..i don't have any loading gif or any indicator..this can be improved i guess..

oparoz:
### Public

> the thumbnail view is gone and plain list is started as default

You need to copy the link from the Gallery side or install the ShareLinks app to get the Gallery link, otherwise you land on the Files side.

> i can drag and drop the photo..although it does not succeed

Make sure you've given the "editing" permission on the link, but there is indeed a bug. I need to use the public endpoint for the MOVE operation.

### Logged in

> when i try moving a folder f1 to another folder f2. i get a js error "gallery.js?v=22125a1

I can't reproduce. I thought it was because of the short name, but I don't have that issue.
Try to see if you can figure out why. Maybe it's a browser issue?

> when i move photos using breadcrumbs..i don't have any loading gif or any indicator..this can be improved i guess..

Yes, I think it would be a nice improvement to be added in the future

tahaalibra:
@oparoz 

## Public
- I copied the link from the gallery side, but still public starts in list mode
  selection_003

## Logged In

To reproduce the javascript error try this
On the files:
- create a folder named(without quotes) "12345" and add 2 or 3 images
- create another folder  named(without quotes) "New Folder" and add 2 or 3 images
- visit gallery from the menu
- try move New Folder to 12345

i think this behavior is coming when i move to a folder that has a number name

oparoz:
Thanks for the screenshot. Please open a bug report regarding the generated link. You have "short links" enabled somehow and those are not properly converted to Gallery links.

Thanks for the test case. I was able to reproduce it. Strange bug :)

tahaalibra:
i guess the link generation bug is unique to official-drag-and-drop-support branch because it was not on the master or official-upload-support branch..should i still report it?

oparoz:
Ah, thanks for checking. It's because this hasn't been rebased...
I'll push a new version in a minute, with the fixes as well

Issue #51, opened 2015-02-28
Make the slideshow controls work with a light background


oparoz:
The current controls icons are dark grey and that works well with a black background as it's non-intrusive, but as soon as a picture with a white background fills the screen or if we manually switch the background to white to better see dark pictures with a transparent background, then we can't use the controls, as they become invisible.

@jancborchardt, feel free to chime in and discuss possible solutions.
## TODO

All that remains are bugs which need to be fixed
- [x] Clean up the icons
- [x] Add missing icons

jancborchardt:
The solution for this is to give the controls a slight dark faded border around them. If anyone takes this up that would be great.

oparoz:
Anyone? :) @jancborchardt @tomneedham @raghunayyar

jancborchardt:
@oparoz I’m experimenting with giving the control icons a slight white/bright shadow (not too extreme). Will do a pull request later.

oparoz:
Nice, thanks :)

oparoz:
Re-opening so that we don't forget about the remaining bug

![](https://cloud.githubusercontent.com/assets/323220/10113489/34d3d754-63e5-11e5-89dc-6b04adac7709.png)

oparoz:
Also, the download and "brightness" icons don't have the new look.

ghost:
This bug will fall off 8.2 if not resolved by 6 Oct

oparoz:
Moving to 9.0 as nobody will be able to fix this on time.

oparoz:
All that remains are bugs which need to be fixed
- [ ] Clean up the icons
- [ ] Add missing icons

ghost:
@oparoz @jancborchardt totally unclear to me who's working / planning to work on the remaining bugs

oparoz:
ownCloud designers need to fix the icons in core and replace the 2 in Gallery which use fontawesome with custom ownCloud ones.

I've copied the TODO to the OP

ghost:
Not sure who "ownCloud designers" actually is. Sorry if I am being dense about that

oparoz:
@jancborchardt or a member of @owncloud/designers who can fix the remaining problems.

jancborchardt:
@oparoz can you be more specific? I assume:
- the download icon needs a version with shadow
- what’s the issue with that extra shadow on the play icon you posted above? Seems like a JS or CSS issue rather than an icon issue, with loading the pause icon partially?
- what is the brightness icon?

[...]

PR #533, opened 2016-02-25
update toggle icon, remove outdated download icon, ref #51

jancborchardt:
Fixing https://github.com/owncloud/gallery/issues/51 together with https://github.com/owncloud/core/pull/22654

Please review @owncloud/designers @oparoz

@oparoz I removed the outdated download icon because view-download from core should be used. I didn’t add the toggle icon to core since we should still think about a better solution here than a button. Anyway, different discussion. :)
### Tested on
- [x] Windows/Firefox
- [x] Windows/Chrome
- [ ] Android 4.x/Chrome
- [x] iOS9/Safari
- [x] BlackBerry10/Browser 
### Check list
- [ ] Commits have been squashed
### Reviewers

@MorrisJobke @setnes @demattin


MorrisJobke:
:+1:

oparoz:
Thanks for the fixes :)

slideshow.css needs to be updated to use the icon from core since it doesn't exist in Gallery any more.

jancborchardt:
@oparoz for some reason the download icon does not show up for me in the detail view. How can I invoke it?

Because otherwise I would just remove the line `background-image: url("../img/download.svg");`

oparoz:
@jancborchardt - Maybe you're not on the right branch in `core` since it needs to be available there?

jancborchardt:
@oparoz I’m on the icon-fixes branch in both core and gallery. This is what I see:
capture du 2016-02-25 21-54-27
The download icon is neither next to the file name nor at the top. What am I doing wrong? :D

oparoz:
Must be a cache, issue, I've just tried with `core` master since it now has the new icon and disabling `background-image` in `slideshow.css` does the trick.

jancborchardt:
@oparoz updated the CSS. Still can’t see the icons :\

Actually, are those full blocks of CSS not outdated now, maybe except setting the height / width (although that should be 32px, no?)? Because the size of the image is the same now as the close / previous / play buttons, and should be displayed the same.
Sorry I can’t fix it myself … no idea why that doesn’t work.

oparoz:
> updated the CSS. Still can’t see the icons :\

Hmmm... That's going to be a problem if it doesn't work for everybody...
Anything in the network tab? What does the CSS for the icon look like? There should be a background-image entry.
Maybe you're using a different theme?

> Actually, are those full blocks of CSS not outdated now, maybe except setting the height / width (although that should be 32px, no?)? Because the size of the image is the same now as the close / previous / play buttons, and should be displayed the same.

Yes, I played with that yesterday and they should be 32px I can fix that in this PR. I was also thinking of moving the play button next to the others for consistency, as a first step towards having some sort of menu which can shrink to 3dots.

oparoz:
@jancborchardt - You need to launch the slideshow from Gallery. There is no download icon on the Files side since there is already one on every row.

jancborchardt:
@oparoz I also launched it from the gallery side, especially in the screenshot I took above. Besides, the actions should always be there in the details view, no matter if coming from the Files or Gallery view.

That’s the point – sure they are in the list, but if I see a picture I want to share or download, I want to do it right then and there. Not need to go back and find it in the list and then share it. ;)

oparoz:
> ..., especially in the screenshot I took above

I doubt that. The URL in the screenshot clearly shows a public Files page ;)

> That’s the point – sure they are in the list, but if I see a picture I want to share or download, I want to do it right then and there. Not need to go back and find it in the list and then share it. ;)

Share, yes, especially on social networks, but for downloads, you need to build the case in a feature request because it's not that straight forward. I find it much more useful to be able to pick a few pictures and download them in one go than to have to download every single one of them.

jancborchardt:
Merging as we can fix anything else going forward.

PR #552, opened 2016-03-02
Update the env.global.CORE_BRANCH setting in .travis.yml

nickvergessen:


LukasReschke:
:+1:

DeepDiver1975:
@oparoz please check the failing tests on travis - since I have no clue about codeception I cannot be of any help :cry:

oparoz:
All good, the error happens because stable9 is slower than stable8.2 in loading Files, so I'll probably have to make the test sleep for a while before testing for the presence of Files. It usually only happens with PostgreSQL.

PR #558, opened 2016-03-04
Close the share dialogue when performing another action, fixes #545

### LINKED ISSUE #545: Close the share dialogue when performing another action
### Steps to reproduce
1. Open the share dialogue
2. Click on an album

or
1. Open the share dialogue
2. Click on the info button
### Expected behaviour

The share dialogue should close
### Actual behaviour

The share dialogue stay open

virajparimi:
Fixes: #545
Licence: MIT or AGPL
### Description

Close the share dialogue when performing another action like opening album or showing infobar.
### Tested on

-[ ]Ubuntu/Chrome
### Reviewers

@oparoz 
@raghunayyar

### DIFF
--- a/js/galleryalbum.js
+++ b/js/galleryalbum.js
@@ -157,6 +157,7 @@
 			event.stopPropagation();
 			// show loading animation
 			this.loader.show();
+			$('#dropdown').slideUp();
 		},
 
 		/**
--- a/js/galleryinfobox.js
+++ b/js/galleryinfobox.js
@@ -26,6 +26,7 @@
 		 * Shows an information box to the user
 		 */
 		showInfo: function () {
+			$('#dropdown').slideUp();
 			if (this.infoContentContainer.is(':visible')) {
 				this.infoContentContainer.slideUp();
 			} else {
--- a/js/public.js
+++ b/js/public.js
@@ -1,7 +1,0 @@
-$('#remote_address').on("change keyup paste", function() {
- 			if ($(this).val() === '') {
- 				$('#save-button-confirm').prop('disabled', true);
- 			} else {
- 				$('#save-button-confirm').prop('disabled', false);
- 			}
-});
--- a/js/public.js
+++ b/js/public.js
@@ -1,0 +1,7 @@
+$('#remote_address').on("change keyup paste", function() {
+ 			if ($(this).val() === '') {
+ 				$('#save-button-confirm').prop('disabled', true);
+ 			} else {
+ 				$('#save-button-confirm').prop('disabled', false);
+ 			}
+});
--- a/templates/public.php
+++ b/templates/public.php
@@ -19,7 +19,6 @@ script(
 		'galleryrow',
 		'galleryimage',
 		'thumbnail',
-		'public',
 		'vendor/modified-eventsource-polyfill/eventsource-polyfill',
 		'eventsource',
 		'vendor/commonmark/dist/commonmark.min',
@@ -83,7 +82,7 @@ style(
 										<input type="text" id="remote_address"
 											   placeholder="example.com/owncloud"/>
 										<button id="save-button-confirm"
-												class="icon-confirm svg" disabled></button>
+												class="icon-confirm svg"></button>
 									</form>
 								</span>
 				<?php } ?>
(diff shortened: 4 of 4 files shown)


oparoz:
Thank you for your contribution.
1. You need to fill the content of the pull request properly. Remove what is not needed, add some steps if it's not obvious from the original issue, etc.
2. I've tested the fix and it works when entering an album, but not when clicking on the info button, so the alternative solution we've discussed is probably what is needed here, if it doesn't create any conflicts.

virajparimi:
@oparoz Please have a look at the fix.

oparoz:
Thanks.

I'm not a fan of having to add it twice. Especially since that means it would need to be added to every click action in Gallery.

How about adding it here?
https://github.com/owncloud/gallery/blob/master/js/app.js#L41-L43

viraj96:
That does fix the original problem but doesn't fix the one where share dialog  does not slide up when info button is clicked.

virajparimi:
I understand but when I did add there it didn't seem to fix the problem at all.

oparoz:
You're right and on top of that it makes the dialogue unsuable. So it seems this has to be added everywhere. I wished there was a better solution...

Any particular reason you decided not to use `hideDropDown` as originally discussed?

virajparimi:
Sorry, did not mean to add the fix here. Wanted to create another pull request. Anyway @oparoz please check the new fix. And to answer your question I did think to use hideDropDown but seeing the simplicity of the method we discussed later, I thought of continuing with the second approach.

oparoz:
The thing I'm worried about is that `hideDropDown` is doing more than just sliding the dialogue up and this may lead to issues when generating links or sharing as maybe the reference to the current album isn't updated, etc.

Regarding the extra commit, do you know how to undo it? If not I'll explain.

virajparimi:
Please tell, it would help and reduce my time researching.

oparoz:
No problem.

### Recovering your work
1. Fork master into a new branch for the new PR
2. Cherry pick your last commit from this PR
3. Push the new branch
4. Create the PR

### Removing the last commit
1. Go to this PR
2. Do a soft reset with git. Set it to `HEAD~1`. That means HEAD minus 1 commit.
3. Revert your changes locally
4. Do a forced push

oparoz:
This still need to be finished. I'll feel safer if hideDropDown was used.

virajparimi:
@oparoz Do you want me to replace slideUp function with hideDropDown, as using that also works. But you did not want to change two files, right?

oparoz:
> Do you want me to replace slideUp function with hideDropDown, as using that also works

Yes, please.

> But you did not want to change two files, right?

I think that's unavoidable.

virajparimi:
@oparoz I have noticed that even after clicking on the image when slideshow opens, the share dialog does not slide up. I have fixed that by adding hideDropDown in slideshow.js file, but I think I found a bug too while doing this. Suppose, you click on image and slideshow starts but later press esc key, then again click on image, image does not show up. It just shows loading icon. Do you also get the same error?

oparoz:
> I have noticed that even after clicking on the image when slideshow opens, the share dialog does not slide up

Good catch. That's one reason I wanted to have a universal closer. Can you check how it works in Files? Per example the 3-dots menu slides-up when we click elsewhere in the interface, so maybe it listens for such events?

> Suppose, you click on image and slideshow starts but later press esc key, then again click on image, images doe not show up. It just shows loading icon. Do you also get the same error?

I can't reproduce.
1. click on image
2. press esc
3. click on same image or other image

It always loads fine for me.

virajparimi:
> Can you check how it works in Files? Per example the 3-dots menu slides-up when we click elsewhere in the interface, so maybe it listens for such events?

I think they have used `OC.hideMenus()` function in different files, so maybe they are doing the same thing as well.

oparoz:
Hehe, OK, let's do it the same way then. Please also close the infoButton when launching the slideshow.

virajparimi:
@oparoz Would you mind if I ask you to close this PR so that I can send another one from a different branch?

oparoz:
It's not a problem if you do it yourself, just make sure to copy the information, add the labels and notify people if it's ready to review.

virajparimi:
Ok, got it. Thanks!

Issue #554, opened 2016-03-03
Cannot download shared albums that were shared via link


davitol:
### Steps to reproduce
1. Go to Gallery and browse into an album
2. Share it via link
3. Copy the link a nd paste it
4. Try to download the album
### Expected behaviour

It should be possible to download the album. 
### Actual behaviour

An error is shown. File not found... (see screenshot below)
### Server configuration

**Operating system**: ubuntu 14.04

**Web server:** apache 2.4.7

**Database:** mysql

**PHP version:** 5.5.9

**ownCloud version: ownCloud Enterprise Edition 9.0.0 RC1 (testing)
### Logs

No logs in owncloud.log or in web console

**Updated from an older ownCloud or fresh install:** Fresh install

**Are you using external storage, if yes which one:** no.
**Are you using encryption:** no
**Are you using an external user-backend, if yes which one:** no
### Client configuration

**Browser:** Firefox for Mac OSX

![screen shot 2016-03-03 at 15 50 11](https://cloud.githubusercontent.com/assets/10596560/13498078/b2930638-e158-11e5-9e9f-d7a835bab577.png)

![screen shot 2016-03-03 at 15 50 01](https://cloud.githubusercontent.com/assets/10596560/13498101/d2137dd0-e158-11e5-9622-29524cc608c5.png)

@rullzer @oparoz

oparoz:
Confirmed. We have to remove the folder name from the path or get `core` to fix the download endpoint to use the path to generate the name of the archive.

oparoz:
https://github.com/owncloud/core/issues/22836, if accepted is probably going to be for 9.1, so the best thing to do is to just remove the top folder from the path or use another variable which gives us the correct path.

oparoz:
Re-opening as only 9.1 has been merged so far.

oparoz:
Fix has been added to stable9

PR #561, opened 2016-03-05
#554: Shared Albums can now be downloaded via shared link

imjalpreet:
Fixes: #554 
### Description

There was an error while trying to download when tried through the link shared from inside an album in the gallery. This has been solved now.
### Features

Shared Albums can now be downloaded even from the link shared from inside an album in the gallery.
### Tested on
- [x] Ubuntu 14.04/Firefox
- [x] Ubuntu 14.04/Chrome
### Reviewers

@oparoz Please have a look at this. I think I have solved #554. Thanks!

### DIFF
--- a/js/galleryutility.js
+++ b/js/galleryutility.js
@@ -170,7 +170,7 @@ window.Gallery = window.Gallery || {};
 
 			if (Gallery.token) {
 				params.token = Gallery.token;
-				subUrl = 's/{token}/download?path={path}&files={files}';
+				subUrl = 's/{token}/download?dir={path}&files={files}';
 			} else {
 				subUrl = 'apps/files/ajax/download.php?dir={path}&files={files}';
 			}


oparoz:
Thanks for trying to fix this, but unfortunately, this is not it ;)

There are 2 public landing pages in ownCloud, one on the Files side and one of the Gallery side.

Your fix redirects users to the Files side, but we want to keep them on the Gallery side.

What needs to happen is to modify [`buildFilesUrl`](https://github.com/owncloud/gallery/blob/master/js/galleryutility.js#L164-L179) so that it works in all cases:
1. Downloading a single file
2. Downloading the current album

imjalpreet:
@oparoz I will try to fix this as soon as possible. Thanks for the feedback.

imjalpreet:
@oparoz I have tried to remove the bug. Can you please give some feedback. Thanks!

oparoz:
Seems to work for me, good job.

Can you see if it's possible to URL decode the path? Because any folder name with spaces and non ascii characters will be encoded and it makes the zip ~~file~~ filename unreadable.

imjalpreet:
@oparoz Sorry for the delay. I tried folder name with spaces but the zip file is still readable. Can you please tell me whether I am missing out something? Thanks!

oparoz:
Here is an example filename if we keep it as is:
brand%20guideline%20logos.zip

The `%20` make the name unreadable and it's worse with things like accents.

And don't worry about the delay. There is no deadline for this ;)

imjalpreet:
@oparoz Ah ohk. I first thought you meant that zip file is becoming corrupt due to that. I also saw this earlier. Now, I will try to fix this issue. :+1:

oparoz:
I've just noticed that if we're at the top folder, the zip is named download.zip. I think it would be good to get the name there.

I'm waiting for this https://github.com/owncloud/core/pull/22902/files to be merged into stable9 to see if that's the correct behaviour and, if yes, then something similar should be implemented.

imjalpreet:
@oparoz I also observed that. Right now I am trying to fix that and the name when an album is downloaded. I have fixed the name when an Image is downloaded but I am still trying to figure out how to fix it for albums. 
Secondly, if we are in the list mode in the shared link then **download.zip** is not only used for when we are in the top folder but anywhere. I think that should be fixed as well.

oparoz:
> Secondly, if we are in the list mode in the shared link then download.zip is not only used for when we are in the top folder but anywhere. I think that should be fixed as well.

That's the fix I linked to in my previous message

imjalpreet:
@oparoz Oh Ohk! 
I have just committed the fix for file name of images. Can you please give some feedback? Thanks! :smile:

oparoz:
OK, that didn't work, but looking at the code, it's no surprise, the changes have nothing to do with the name of the folder to download.

Can you explain what you were trying to achieve with the changes in files.php?

imjalpreet:
@oparoz I think you misunderstood what I am trying to do. The code that I have committed is for the filename of an image(i.e. if you try to download only an image) and not the folder, I am still trying to fix it for the folder name.

oparoz:
Thanks for clarifying, but that change is not necessary as filenames are correct when downloading individual files.
For the folder name, it could be that it's not possible to fix it ourselves and that's it's already fixed with that latest patch. I'm going to try that patch now.

oparoz:
Yeah, the same bug is present in Files, so that endpoint needs fixing. We're sending a proper filename and we get an encoded one.

oparoz:
So basically, your initial fix should be enough, the rest should be fixed in `core`. Unfortunately, this will probably only be available in 9.0+.

I'd say, revert to the first fix and I'll merge that, then you can create a backport for stable9

OK?

Also, this is a good example of how things work when hacking Gallery. There are some dependencies with some `core` endpoints or even Javascript files and it's usually best to try and adapt, unless it becomes too much of a problem, at which point it's always possible to include the functionality in Gallery.

imjalpreet:
Yeah, I also checked it with that patch, the folder names have started to come but they are still encoded and we need to figure out a way to correct them. 
I have two questions:
One is that do we need the code for file names that I had committed in this PR?
Second is that can you please help me by telling me where exactly are we sending the name of the folder? I am not able to figure out that from a long time. 
Thanks!

imjalpreet:
Sorry, I missed your reply. So you want me to remove my last commit, right?
I have a small question: What do you mean by "create a backport for stable9"?
I will slowly get used to all these dependencies between core and gallery.
Thanks!

oparoz:
> One is that do we need the code for file names that I had committed in this PR?

No, as they are correct when using the Gallery endpoint afaik

> Second is that can you please help me by telling me where exactly are we sending the name of the folder?

In this method, the name of the folder is `files` and "path" contains the top, shared folder.

If you want to rename `files` in the method's signature, I'm OK with it.

> So you want me to remove my last commit, right?

You only have one commit, so do it like you usually do and make it so that there is only one commit containing your fix.

> What do you mean by "create a backport for stable9"?

All commits are merged against master, which matches the branch in `core`. So important fixes should be backported to older branches. 
In Gallery, I only do it for very important change as I don't have enough time to allocate to that.
I think this fix is important, because getting a 404 when trying to download a folder offers a poor UX.

oparoz:
To backport, you simply fork the stable branch you need to use and cherry pick the commit to that fork.
when creating the PR, don't forget to select the stable branch and not master.

Also indicate in the PR name the name of the branch, as well as the original PR.

imjalpreet:
@oparoz I have reverted my commit.
In the mean time, I am trying to figure out what backport is.

imjalpreet:
@oparoz I wanted to clarify something about backport:
Do you want me to create the same PR to the stable9 branch?
Which branch are you referring when you say that include the branch name in the PR? Do you mean stable9?
Thanks!

oparoz:
> Do you want me to create the same PR to the stable9 branch?

Exactly

> Which branch are you referring when you say that include the branch name in the PR? Do you mean stable9?

Yes: `[stable9] Shared Albums can now be downloaded via shared link, backport of #561`

I think there is some instruction about that in the `core` wiki, but if you look at the list of PRs in `core` that should give you an idea.

oparoz:
:+1: for this fix

Issue #556, opened 2016-03-03
Do not allow empty urls to be submitted in "add to your owncloud"


oparoz:
Originally reported by @jancborchardt 
### Steps to reproduce
1. Go to a public Gallery
2. Click on "Add to ownCloud"
3. Press "Enter" or the confirm button **without entering any url**
### Actual behaviour

![federated-sharing-no-url-put](https://cloud.githubusercontent.com/assets/925062/10111548/ef16a342-6428-11e5-8bbc-261a37ddabbf.png)
### Expected behaviour

When no URL is put in, nothing should happen (the button should be disabled).

The fix is here and needs to be ported to Gallery:
https://github.com/owncloud/core/commit/08fd4b38f895a5e9a91cdc476103c26be68dda80

virajparimi:
I would like to work on this one. But I am not able to reproduce the error. Where do we enter the url, do we do it while sharing the file when share dialog opens?

kavishdahekar:
@viraj96 I think it is already fixed. 

> The fix is here and needs to be ported to Gallery:
> owncloud/core@08fd4b3

prastut:
@oparoz This seems to be fixed.

oparoz:
In order to understand the problem, you need to read this:
https://doc.owncloud.org/server/9.0/admin_manual/configuration_files/federated_cloud_sharing_configuration.html#creating-federation-shares-via-public-link-share

and then try to apply what's described in the OP **from the Gallery side**. You'll experience the same bug and will need to port the fix from `core` to Gallery.
In order to fix the problem, you need to have access to 2 different owncloud instances, to make sure that the connection is successfully established when entering a valid URL.

oparoz:
I've updated the original report to make it clearer, since on the `core` side, people are more experienced with these kind of issues and can usually read between the line. That's probably not the case for people new to the project.

tahaalibra:
i added patch using 08fd4b3.
by the way just wanted to know..can it also be done by adding  attribute 'required' to the 'input' tag.

oparoz:
@tahaalibra - I don't know if you're familiar with Github, but on the right, you can see that the task was already assigned. I can only do it with registered ownCloud members, so it's also a good idea to also check the list of PRs before submitting a new one. That way there is no duplication of effort :)

tahaalibra:
@oparoz i see..somehow i missed to see if it was assigned...i will surely check carefully next time..
No harm Done..i ended learning more about the code...which was my original intention

[...]

PR #562, opened 2016-03-06
Do not allow empty urls to be submitted in "add to your owncloud"

virajparimi:
Fixes: #556 

Licence: AGPL
### Description

Fixed the error that came when no URL was entered. Added a new public method to galleryview.js file.
### Tested on
- [x] Ubuntu 14.04 LTS/Chrome
### Reviewers

 @oparoz

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -40,6 +40,13 @@ body {
 	opacity: .5;
 }
 
+#save-button-confirm:disabled, #save-button-confirm:disabled:hover, #save-button-confirm:disabled:focus {
+    -ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=20)";
+    filter: alpha(opacity=20);
+    opacity: .2;
+    cursor: default;
+}
+
 #save-button-confirm:hover,
 #save-button-confirm:focus {
 	opacity: 1;
--- a/js/galleryview.js
+++ b/js/galleryview.js
@@ -35,6 +35,7 @@
 			// Only do it when the app is initialised
 			if (this.requestId === -1) {
 				this._initButtons();
+				this._blankUrl();
 			}
 			if ($.isEmptyObject(Gallery.imageMap)) {
 				this.clear();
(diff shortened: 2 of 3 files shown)


virajparimi:
@oparoz  Please have a look at the fix.

oparoz:
Thanks for the PR. There are a few issues with it.
1. A PR has to be based off `master`. This one seems to be a fork of your last PR
2. You have to make a choice in the license field in the PR. If you haven't signed the contributor agreement and you're OK with releasing your work under the MIT license, pick MIT
3. There is no need to include public.js at this stage. All the functions related to the public side are in gallery.js. Maybe this can be split later so that some code is only loaded on the public side, but it's best to move everything at the same time.

virajparimi:
OK 
So, I just need to remove public.js and and copy the same code to gallery.js. And where do I find the contributor agreement?

oparoz:
https://owncloud.org/contribute/agreement/

virajparimi:
@oparoz Have a look at the new one. Have removed the  public.js file.

oparoz:
Your fix doesn't work properly.
You've created a new method which is never called. 
I was wrong before, since this only makes changes to the view, it's best to move it to galleryview, but make sure it's called during the initialisation of the view.

virajparimi:
@oparoz I have fixed the issue which was unresolved. Please have a look.

oparoz:
Thanks. If you leave it where it is in `_init`, then this will be called every time the view is initialised.
The line below your change mentions that there is a special method for initialising buttons only once.
I would add the call there.

Makes sense?

virajparimi:
@oparoz Have made the change you suggested.

oparoz:
A couple of points:
1. We use camelCase to name methods
2. private methods start with an underscore
3. methods should be documented
4. if possible, try to find a meaningful name for the method without making it too long. That makes it easier to understand what it does.

virajparimi:
@oparoz Have a look at  the update. I have made the method public. Thanks!

oparoz:
Why public? It's not being called anywhere but in galleryview

virajparimi:
@oparoz  I initially did make it private but later thought  that maybe later it will be called in some other files so I ended up  making it public. Well now I have updated it. Have a look. Thanks.

oparoz:
I've just tested the solution and there is one more thing missing. The CSS needs to be adjusted to make it work like on the Files side. Without it, we don't really see when the button is enabled/disabled.

virajparimi:
@oparoz I have fixed that issue as well and even squashed all commits. Please have a look. Thanks!

oparoz:
Looks good. I will test as ASAP.

@tahaalibra you can test this and see if it works like on the Files side

tahaalibra:
:+1:  works well
tested on both files and gallery

oparoz:
Thanks. I've tested as well and it works as expected. Thank you @viraj96

PR #572, opened 2016-03-09
empty page message is still too low FIX

tahaalibra:
Fixes: _#569_

Licence: AGPL
### Description

Fix for empty page message, which was coming too low
reduced padding-bottom on div#gallery.hascontrols
### Features

Does not alter any features
### Screenshots

gallery owncloud
files owncloud
### Tested on
- [x] Ubuntu14.04/Chrome
### Public Gallery
- does not get effected...working properly
#### Empty

it is working ok...the message is shifted up in the center...just as in files
#### Half Full

it is working ok...difference is not much noticeable
#### Full

it is working ok...difference is not much noticeable  
### Reviewers

@oparoz

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -72,7 +72,7 @@ div.crumb.last a {
 	overflow: hidden;
 	text-align: justify;
 	top: 45px;
-	padding-bottom: 100px;
+	padding-bottom: 10px;
 	margin-bottom: 45px;
 	margin-top: 0;
 }


oparoz:
Thanks for the fix.
A couple of points:
- Try to describe how your fix works in your description. It helps people quickly understand what's going on
- Add a test plan. For this fix per example, there are lots of scenarios to test: public/logged-in, empty/half-full/full. An example: https://github.com/owncloud/core/pull/19611#issuecomment-146163812

Now about the fix:
It seems to work well.
- The public side overrides that property and sets it at 0. The footer is added below it
- On the logged-in side, there is no footer and the wall ends at the bottom of the page

In general, I don't have a problem with the approach, but if we add a menu to images, then we'll have to grow the bottom to make room.
What do you think @jancborchardt @raghunayyar @tomneedham ?

tahaalibra:
@oparoz yea..this is a basic approach for the problem
i tried improving my comment..i will learn more from the previous pull requests..
i really appreciate you taking your time for giving me useful tips

oparoz:
So @jancborchardt @raghunayyar @tomneedham Is it OK to remove that extra 90 pixels at the bottom of Gallery?

oparoz:
Works for me and can be modified later if there is a need for that empty row at the bottom.

PR #520, opened 2016-02-13
Add uploading support for registered users

oparoz:
This is my proposal for #25. It replaces #485.

This is the drag and drop component and the [ + ] button will be implemented in a separate PR.

This currently re-use the Files' javascript and will switch to webdav once uploading is supported.

I'm looking forward to your feedback :)

---

@karlitschek @jancborchardt @DeepDiver1975 @LukasReschke @PVince81 @MorrisJobke @jospoortvliet @setnes @demattin @arkascha @elpraga @sualko @bugsbane @mmattel

PR #574, opened 2016-03-09
Add uploading support for registered users

### LINKED ISSUE #25: Upload straight from the app [$250]
Being able to upload files is just one click away with the "switch to files view" button, but it could be even better if we could just ~~create folders and~~ upload files straight from the app.
## Specifications

The idea is to be able to easily drag and drop files in the current album, to be able to detect conflicts and to refresh the view once the process is complete.
- [x] Upload via drag and drop ~~only~~
- [x] Upload to current album only. The whole photowall is the drop zone
- [x] Upload as a logged in user only
- [x] Upload both files and folders (Chrome only)
- [x] Show progress bar
- [x] Reload view at end of the process
- [x] Conflict management (same as in Files)
- [x] Don't block in-app navigation. Keep uploading where the upload started (same as in Files)
- [x] Leaving the app kills the upload process (same as in Files)
- [x] Let users upload by clicking on a button in the controls bar

conflict
### Bugs
- [x] Error 500 when uploading on an empty Gallery...
- [x] Conflict dialogue breaks when comparing images located in folders which don't appear in the album representation
## <bountysource-plugin>

Did you help close this issue? Go claim the **[$250 bounty](https://www.bountysource.com/issues/8579594-upload-straight-from-the-app?utm_campaign=plugin&utm_content=tracker%2F9328526&utm_medium=issues&utm_source=github)** on [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F9328526&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oparoz:
Fixes #25. It replaces #485 and #520 .
### Featuring
- Drag and drop component
- Folder uploading on Chrome
- File comparison dialogue
- [ + ] button 
- Improved empty page, now with error messages
### Todo
- [x] Defer quota check in web UI when overwriting shared file owncloud/core@ceaefc2
### Test plan

**Base**
- Uploading file(s) via drag and drop
- Uploading a folder via drag and drop
- Uploading file(s) via the [+] menu

**Additional tests**
- Uploading from an empty Gallery should work and the UI should adapt to the content of the album
- Uploading the same file twice should bring up the conflict dialogue
### Tech review
- The javascript used in `files` is loaded as-is. A special file has been created to replace FileList and Files methods called by those scripts
- `newfilemenu.js` has been imported and modified to make it work with Gallery. We don't need more than the upload entry
- `OC.Upload._isReceivedSharedFile` is replaced with a method which does the lookup in the cache instead of in the HTML
- Some methods were moved from Gallery to GalleryView as it made more sense to have them there
- GalleryImage has 2 new properties: size and sharedWithUser
- The landing page for empty albums has been modified to take into consideration the context in which that page is called and can print the error message, if communicated to the frontend

_Note: Ideally, we should switch to webdav once uploading is supported and re-usable components for the file dialogue move outside of `files`._

---

@karlitschek @jancborchardt @DeepDiver1975 @LukasReschke @PVince81 @MorrisJobke @jospoortvliet @setnes @demattin @arkascha @elpraga @sualko @bugsbane @mmattel @tahaalibra @viraj96 @imjalpreet @mbtamuli @rahulgoyal030

### DIFF
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -82,7 +82,10 @@ class PageController extends Controller {
 		$appName = $this->appName;
 
 		// Parameters sent to the template
-		$params = ['appName' => $appName];
+		$params = [
+			'appName' => $appName,
+			'uploadUrl' => $this->urlGenerator->linkTo('files', 'ajax/upload.php')
+		];
 
 		// Will render the page using the template found in templates/index.php
 		$response = new TemplateResponse($appName, 'index', $params);
--- a/css/mobile.css
+++ b/css/mobile.css
@@ -5,4 +5,9 @@
 		margin-right: 0 !important;
 	}
 
+	/* shorten elements for mobile */
+	#uploadprogressbar {
+		width: 50px;
+	}
+
 }
--- a/css/styles.css
+++ b/css/styles.css
@@ -286,3 +286,16 @@ http://www.bypeople.com/author/comatosed/*/
 .icon-gallery {
 	background-image: url('../img/gallery-dark.svg');
 }
+
+.oc-dialog .fileexists .icon {
+	background-position: center center;
+	background-size: cover !important;
+}
+
+#uploadprogressbar {
+	margin-top: 5px;
+}
+
+.stop.icon-close {
+	margin-top: 5px;
+}
--- a/css/styles.css
+++ b/css/styles.css
@@ -286,16 +286,3 @@ http://www.bypeople.com/author/comatosed/*/
 .icon-gallery {
 	background-image: url('../img/gallery-dark.svg');
 }
-
-.oc-dialog .fileexists .icon {
-	background-position: center center;
-	background-size: cover !important;
-}
-
-#uploadprogressbar {
-	margin-top: 5px;
-}
-
-.stop.icon-close {
-	margin-top: 5px;
-}
--- a/css/upload.css
+++ b/css/upload.css
@@ -1,0 +1,95 @@
+/* Uploading */
+.oc-dialog .fileexists .icon {
+	background-position: center center;
+	background-size: cover !important;
+}
+
+#uploadprogressbar {
+	margin-top: 5px;
+}
+
+.stop.icon-close {
+	margin-top: 5px;
+}
+
+.actions {
+	float: left;
+}
+
+.actions input, .actions button, .actions .button {
+	margin: 0;
+	float: left;
+}
+
+.actions .button a {
+	color: #555;
+}
+
+.actions .button a:hover,
+.actions .button a:focus,
+.actions .button a:active {
+	color: #333;
+}
+
+.actions.creatable {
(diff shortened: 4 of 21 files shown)


tahaalibra:
**Working**
- [x] Uploading file(s) via drag and drop
- [x] Uploading a folder via drag and drop
- [x] Uploading file(s) via the [+] menu
- [x] Uploading from an empty Gallery
- [x] UI is adapting to the changes
- [x] Uploading Same files gives the warning

great  :+1: ...everything seems to be working fine

I will try some more test and take a look at the code..

imjalpreet:
@oparoz Everything seems to be working! :+1: 
Except a few points that I will be mentioning at the end.
I have tested the following:
- Drag and Drop Upload for both file(s) and folder.
- Upload file(s) via [+] button.
- Uploading from an empty Gallery is working (and the UI adapts to the images and also to window resize)
- Conflict Dialogue comes if same files are uploaded more than once.
- Error Messages

But I wanted to clarify one thing, if we try to upload a folder and even if there is one empty file, the entire folder doesn't get uploaded. I think it should not be like this. It think we should try to atleast upload the non empty files and folders. 

Secondly, the error message that is displayed in the above case should be changed. Currently, the error message is 'Error uploading file "Untitled Document": Unable to upload Untitled Document as it is a directory or has 0 bytes'. Since, directory upload is now possible, we should not print in the error message that the reason of not uploading is that it is a directory. I am attaching a screenshot of the following.

One more thing we can do is automatically hide the error message after a certain time instead of keeping it till someone clicks on 'x'.
errormessage

tahaalibra:
I am getting javascript error when viewing a public gallery..it is preventing gallery top open
''galleryview.js?v=95a8dd6…:276 Uncaught TypeError: Cannot set property '_isReceivedSharedFile' of undefined"
selection_001

@oparoz

imjalpreet:
@oparoz Now public gallery is also working fine! :+1:

oparoz:
> I am getting javascript error when viewing a public gallery..it is preventing gallery top open
> ''galleryview.js?v=95a8dd6…:276 Uncaught TypeError: Cannot set property '_isReceivedSharedFile' of undefined"

@tahaalibra - Ah, yes. I've changed approach at the 11th hour and failed to verify that it still worked on the public side. That side doesn't support uploading yet. It's now fixed.

@imjalpreet - All the problems you've described should happen on the Files side as well and some reports have probably already been opened in `core`. Go see if that's the case and if not, add your suggestions. Some of these issues are not trivial to fix, but if you have solutions, `core` is also accepting PRs :)

imjalpreet:
@oparoz Yeah Sure. I will have a look in core.

tahaalibra:
@oparoz now working on public as well as logged in !!!!!

jospoortvliet:
Awesome work @oparoz !

Issue #573, opened 2016-03-09
Only show sort buttons if there is more than 1 image in the album


oparoz:
## Feature request

**User type**: All

**User level**: All
### Description

It doesn't make sense to show sort buttons if there is only a single image in the album
### Benefit / value

- Make the app look cleverer
- Give a task to someone who wants to learn more about the inner-working of the app
### Risk / caveats

Some users get confused when the interface changes

virajparimi:
I would like too work on this one. Will get back to you with progress.

virajparimi:
@oparoz I have been able to solve the problem when there is only one image, but I have one doubt. Suppose that we have only one folder in the gallery then also should the sort buttons appear?

oparoz:
@viraj96 - It's the same problem. The buttons should not show when there is only one album

oparoz:
There is a regression:
1. Add pictures to a folder
2. Create a sub-folder
3. Enter the sub-folder
4. Add a `.nomedia` file
5. Switch to Gallery
6. Press on the parent folder in the breadcrumb

```
TypeError: album.images is undefined  in galleryview.js
```

@viraj96 - You need to make sure these have been defined before trying to use them.

virajparimi:
Ok. Will look into it and fix it. By the way, how do I talk to you about gsoc - 2016 program, like where are you available?

oparoz:
I've added my contact details to the Gallery wiki

virajparimi:
@oparoz I am not getting such error but I am getting something like this. 
![owncloud_1](https://cloud.githubusercontent.com/assets/14194935/13754798/53245a78-ea3e-11e5-82bc-3fea4b85ff67.png)
Are you referring to the same thing? Else could we just put something like `if(typeof album.image!==undefined)` then do stuff else don't use that.

oparoz:
I think you've missed a step, the error shows up when you click on the parent folder.
But yes, the fix would be the same and please use underscore to check if it's undefined.
Thanks!

virajparimi:
@oparoz Some doubts. Can `album.subAlbums` also be undefined ever? And suppose if it is, what would be our sum if both `album.images` and `album.subAlbums` both become undefined? Would it be 0.

oparoz:
> Can album.subAlbums also be undefined ever?

I think so, you need to check the flow of the app. When does it make that test and when does it get a list of images.
It could be that there is a very similar problem when opening gallery for the first time and it's empty.

[...]

PR #577, opened 2016-03-10
Only show sort buttons if there is more than 1 image in the album.

virajparimi:
Fixes: #573 

Licence: AGPL
### Description

Fixes the problem and shows sort button if more than 1 image in album.
### Features

Works for images as well as folders.
### Tested on
- [X] Ubuntu 14.04/Chrome
### Reviewers

@oparoz

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -40,6 +40,13 @@ body {
 	opacity: .5;
 }
 
+#save-button-confirm:disabled, #save-button-confirm:disabled:hover, #save-button-confirm:disabled:focus {
+    -ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=20)";
+    filter: alpha(opacity=20);
+    opacity: .2;
+    cursor: default;
+}
+
 #save-button-confirm:hover,
 #save-button-confirm:focus {
 	opacity: 1;
--- a/css/public.css
+++ b/css/public.css
@@ -40,13 +40,6 @@ body {
 	opacity: .5;
 }
 
-#save-button-confirm:disabled, #save-button-confirm:disabled:hover, #save-button-confirm:disabled:focus {
-    -ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=20)";
-    filter: alpha(opacity=20);
-    opacity: .2;
-    cursor: default;
-}
-
 #save-button-confirm:hover,
 #save-button-confirm:focus {
 	opacity: 1;
--- a/js/galleryview.js
+++ b/js/galleryview.js
@@ -35,7 +35,6 @@
 			// Only do it when the app is initialised
 			if (this.requestId === -1) {
 				this._initButtons();
-				this._blankUrl();
 			}
 			if ($.isEmptyObject(Gallery.imageMap)) {
 				this.clear();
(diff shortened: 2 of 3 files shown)


oparoz:
> Fixes the problem and shows sort button if more than 1 image in album.

It would be good if you could describe in more details the logic you've used. That way the principles can be validated before the code.

Looking at the code, I see a couple of issues:
1. The first conditions shows the buttons if there are 0 albums or images
2. The 2nd conditions allows for 2 elements to be shown on screen

So it seems you need to write more tests in your test plan to try and cover more cases

virajparimi:
@oparoz First condition is pretty trivial to fix. But the second condition is for when 1 folder and 1 image is present. At that time should we show sort buttons or not?

oparoz:
> when 1 folder and 1 image is present. At that time should we show sort buttons or not?

Yes. Even though it appears that no sorting would be necessary, because all elements can bee seen at once, I have no way of defining from which point it is.
Should people not have the capability to sort less than 5 elements? 3? 10? I don't think we can arbitrarily pick one.

oparoz:
Ah and I read the 2nd condition wrong. So, the logic is correct there, but try to combine 1 & 2 if you can.

If we have more than one element -> show buttons

virajparimi:
@oparoz Have a look at this one.

oparoz:
That doesn't work if I have 10 images and 0 albums, no? (Add this case to the test plan if it doesn't)

virajparimi:
@oparoz My bad, I missed those conditions. Anyways the new one adds those conditions as well.

oparoz:
OK, now try to combine these conditions into 1, if possible.

virajparimi:
@oparoz Have done. Please check. Thanks!

oparoz:
I think I have a better idea. Add the number of images and albums and if the sum is greater than one, then show the buttons. What do you think?

virajparimi:
@oparoz :+1: Certainly a good idea, as it decreases the number of redundant comparisions and combines them. Have updated the file.

oparoz:
OK about the idea, but have you tested your code? By looking at it, I'd say the way you wrote it doesn't work.

virajparimi:
@oparoz Sorry for the inconvenience.

oparoz:
We all make mistakes :), but that's why I think the test plan is important. It makes it easy to quickly review all cases.

Let me know when you've tested the solution and I'll review it.

virajparimi:
@oparoz I have tested it and as far as I can see there are no errors. I worked with 1 image, then 2 images, similarly for albums and then both of the together.

oparoz:
OK, I'll let you summon some testers ;)

virajparimi:
Ok, @imjalpreet @tahaalibra  Need your reviews about this. What do you think?

tahaalibra:
i think @mbtamuli was working on this. please review...
i will test this and comeback to you

mbtamuli:
Everything is working properly. :+1: 
- 1 Image.
- 1 Folder.
- 1 Image + 1 Folder
- Multiple Images
- Multiple Folders
- Multiple Images + Multiple Folders

oparoz:
Thanks @mbtamuli

Issue #405, opened 2015-10-04
Drag and drop files and folders [$250]


oparoz:
As soon as the app supports [uploading files](https://github.com/owncloud/gallery/issues/25), there will be the need to re-organise them.
## Specifications

Files/folders should be draggable and folders/breadcrumbs should be drop targets. There is no support for multi-element select.
- [x] Make files draggable
- [x] Make folders draggable
- [x] Draggable elements must be resized and made semi-transparent so that users can easily drop them on the targets
- [x] Turn folders into drop zones
- [x] Turn breadcrumb elements into drop zones so that users can move files to parent folders
- [x] Add visual feedback to drop zones
- [ ] ~~Only turn albums with write permissions into drop targets~~ Requires #557
- [x] Move files and folders when dropped onto an album and refresh the view
### FAQ

**Q:** I get a permission denied after dropping files on an album. Why make it a drop target then?
**A:** It's actually not possible today to determine which albums can be real drop targets as we don't have any information about sub-albums beside their name. See #557 for details.
##

oparoz:
I'll post a roadmap soon

oparoz:
@jospoortvliet @jancborchardt @karlitschek @LukasReschke I've posted some specs. Let me know if this meets your requirements

jancborchardt:
Looks good! cc @owncloud/designers for checking as well.

jospoortvliet:
I think it's great!

oparoz:
There is now a solution with https://github.com/owncloud/gallery/pull/519.

I wanted to visually indicate that some folders cannot be drop targets, but I don't think having a bix X across a folder is going to work, especially for the breadcrumb. The next solution would be a pop-up which would appear when the item hovers, but I'm worried that it might be unreadable.
Thoughts?

oparoz:
Added a FAQ

Issue #565, opened 2016-03-08
Add the checkbox in the share dialogue to allow uploading from the public side


oparoz:
## Feature request

**User type**: All

**User level**: All
### Description

Users who want to let people they share a link with upload new files can't do it from the Gallery side
### Benefit / value

Feature parity with the share dialogue in Files
### Risk / caveats

The share dialogue will be rewritten this year, so not too much time should be spent on this as this would be a temporary solution.
### Sponsorship

None

imjalpreet:
@oparoz I have started to work on this as well as #213. Will get back to you with some progress soon.

imjalpreet:
@oparoz Can you please give a little pointer on how can I proceed with this? I have been trying to fix this but I haven't been successful till now. A small pointer from your side would be really helpful. Thanks!
Sorry, I haven't been able to contribute much from yesterday due to my university commitments.

oparoz:
Let's see if that helps :)
- https://github.com/owncloud/gallery/blob/master/js/vendor/owncloud/share.js#L172-L189
- https://github.com/owncloud/core/blob/master/apps/files/templates/list.php#L43
- https://github.com/owncloud/core/blob/c5a200c4198de951dabdd3f783ade655b8ea6434/apps/files/list.php#L28

imjalpreet:
@oparoz I have already headed in this direction. I was also referring to these files only. I am trying to figure out how to do the same in gallery app. Thanks!

oparoz:
- https://github.com/owncloud/gallery/blob/master/templates/part.content.php#L100
- https://github.com/owncloud/gallery/blob/master/controller/pagecontroller.php#L84

oparoz:
Fix is in #587

PR #581, opened 2016-03-10
Official drag and drop support

oparoz:
Solution for #405, replaces #519 
- You can drag images and albums to any album or the breadcrumb
- If successful, the photowall is redrawn
- If you're not allowed to make the change, elements go back to their initial position
### Possible improvements
- ~~Drop targets could be nicer~~ Done
- ~~Dragged albums maybe need a border?~~ After extensive use, I don't think so
### Screencasts

http://screencast-o-matic.com/watch/cDnl0rhWt7

**Screenshot**
dropzones
### Test plan

Name at least one file and one folder with numbers only

As a logged-in user:
- Drag a file to a folder and drop it
- Drag a file to a breadcrumb and drop it
- Drag a folder to another folder and drop it
- Drag a folder to a breadcrumb and drop it
- Make sure sometimes the number folder is the target and sometimes the dragged object
- **Test on as many browsers as you can. Some browsers don't handle authorization the same way**

Do the same on the public side, but make sure you've added the "editing" permission to the link.
### Caveats

Read the FAQ in #405 
### TODO
- [x] Create our own instance of Files.Client so that we can pick the URL to use

---

@karlitschek @LukasReschke @tahaalibra @viraj96 @imjalpreet @mbtamuli @rahulgoyal030


oparoz:
It's ready to review again. Test plan has been updated.
@jancborchardt - Would be good to get your go ahead regarding the dropzones

mbtamuli:
@oparoz All the tests successfully passed! :+1: 
Just one problem on the Gallery side.

## Logged In
- The Dropzones are properly marked.
- Numbered files and Folders can be dragged and dropped.
- Drag and Drop to breadcrumbs are working.

## Public
- From the Gallery side, there was _no_ option to add the "editing" permission to the link. So, **no** drag and drop. @tahaalibra @imjalpreet @viraj96 Can you confirm?

Tested on - 
- [x] Chrome v49/Ubuntu15.04
- [x] Firefox v44/Ubuntu15.04

Two other warnings came up, though maybe not due to this PR.
- `Deprecation warning: tipsy is deprecated. Use tooltip instead.`
- When I click the share link, the first time after loading the page, _only on the Gallery side_ this warning is logged.
  `Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience.`

imjalpreet:
@mbtamuli Yeah, you are right. Right now, there is no option to give editing permission to the link from the gallery side. I am working on that Issue ( #565 ).

oparoz:
Thanks @mbtamuli . The deprecation warning comes from `core` and the synchronous XMLHttpRequest warning is generated by the imported share class. It waits on getting the shares before building the UI. That could be fixed by using a spinner and dynamic resizing like in the infobox.

tahaalibra:
tested on various conditions...working fine...only thing is that no option editing permission and also i can drag and drop files/folder in public view....it fails but i think it would be better to remove the option of drag and drop in public view( when editing permission is not given )

oparoz:
@tahaalibra - Permission checking on the public side shouldn't be too hard to implement for anybody once this is merged :)

Issue #213, opened 2015-07-14
No possibility to send an email in share dialog


sonymaster:
If you open the share dialog from the top right share button, you can share the gallery with OC users or/and as Link. But the possibility to send an email with share infos is not displayed like in the Files app.

oparoz:
You're correct, thanks for notifying us.

oparoz:
@rullzer @PVince81 - Is there something special to do to get the share by email section in the share dropdown?

imjalpreet:
@oparoz Can you please give little more detail regarding this? I am not exactly able to figure out what the problem is. I didn't understand the line : "_**..possibility to send an email with share infos..**_". Thanks!

oparoz:
If you compare the dialogue in Gallery with the one in Files, you'll notice that you cannot send an email in Gallery after having created a share or a public link.
There are screenshots in #498.

Also, in `share.js` I've commented out the code as it's not in use. You need to uncomment the parts relevant to email sharing and update it to make it work like the rest of the class.

oparoz:
This commit should help you see the changes made to the ajax calls: https://github.com/owncloud/gallery/commit/48cf8534c9ff9b8ca6eb88b2720e33651e532e3b#diff-b3a827de53066e0a69e83addf18ddd46

imjalpreet:
@oparoz Thanks! I will come up with a solution as soon as possible. I am trying to understand the flow right now.

oparoz:
That class is a pita... This one is easier to understand because it doesn't have all the UI stuff mixed up:
https://github.com/owncloud/core/blob/master/core/js/shareitemmodel.js

Eventually, we should migrate to that, using Backbone.

imjalpreet:
![error](https://cloud.githubusercontent.com/assets/6432146/13701822/4e15694a-e7b0-11e5-86ae-6a712bcf0d39.png)
@oparoz I wanted to ask you a question. When I am trying to send a mail from files app, I am getting an error message(attached screenshot). Can you tell me do we have to setup mail before this? Thanks!

oparoz:
I don't see the screenshot, but look in the admin settings. There you can authorise emails to be sent.

imjalpreet:
@oparoz No problems. I have figured out how to do it. Thanks!

oparoz:
Ah, good :)

oparoz:
[...]

PR #586, opened 2016-03-11
Issue #213: No possibility to send an email in share dialog

imjalpreet:
Fixes: \* #213 *

Licence: AGPL
### Description and Features

It is now possible to send an email from the share dialog in the Gallery
- [x] Emails containing the public link are sent to the given email address
- [x] Email notifications can be sent to internal users
### Screenshots or screencasts

screenshot2
screenshot1
### Tested on
- [x] Ubuntu 14.04/Firefox
- [x] Ubuntu 14.04/Chrome
### Test Plan
- First step is to setup your email settings in the admin settings.
- Go to the Gallery App.
- Now, try to share a folder via link.
- Try to send the mail to a person by entering the email.
- Try to test in various browsers.
### Reviewers

@oparoz Please have a look at it.

### DIFF
--- a/controller/pagecontroller.php
+++ b/controller/pagecontroller.php
@@ -82,11 +82,7 @@ class PageController extends Controller {
 		$appName = $this->appName;
 
 		// Parameters sent to the template
-		$params = [
-			'appName' => $appName,
-			'uploadUrl' => $this->urlGenerator->linkTo('files', 'ajax/upload.php'),
-			'publicUploadEnabled' => $this->appConfig->getAppValue('core', 'shareapi_allow_public_upload', 'yes')
-		];
+		$params = $this->getIndexParameters($appName);
 
 		// Will render the page using the template found in templates/index.php
 		$response = new TemplateResponse($appName, 'index', $params);
@@ -174,6 +170,35 @@ class PageController extends Controller {
 	 */
 	public function slideshow() {
 		return new TemplateResponse($this->appName, 'slideshow', [], 'blank');
+	}
+
+	/**
+	 * Returns the parameters to be used in the index function
+	 *
+	 * @param $appName
+	 *
+	 * @return array<string,string>
+	 */
+	private function getIndexParameters($appName) {
+
+		// Parameters sent to the index function
+		$params = [
+			'appName' => $appName,
+			'uploadUrl' => $this->urlGenerator->linkTo(
+				'files', 'ajax/upload.php'
+			),
+			'publicUploadEnabled' => $this->appConfig->getAppValue(
+				'core', 'shareapi_allow_public_upload', 'yes'
+			),
+			'mailNotificationEnabled' => $this->appConfig->getAppValue(
+				'core', 'shareapi_allow_mail_notification', 'no'
+			),
+			'mailPublicNotificationEnabled' => $this->appConfig->getAppValue(
+				'core', 'shareapi_allow_public_notification', 'no'
+			)
+		];
+
+		return $params;
 	}
 
 	/**
--- a/js/vendor/owncloud/share.js
+++ b/js/vendor/owncloud/share.js
@@ -254,18 +254,18 @@
 						t('core', 'Allow editing') + '</label>';
 						html += '</div>';
 					}
-					 /*var mailPublicNotificationEnabled = $(
-					 'input:hidden[name=mailPublicNotificationEnabled]').val();
-					 if (mailPublicNotificationEnabled === 'yes') {
-					 html += '<form id="emailPrivateLink">';
-					 html +=
-					 '<input id="email" style="display:none; width:62%;" value="" placeholder="' +
-					 t('core', 'Email link to person') + '" type="text" />';
-					 html +=
-					 '<input id="emailButton" style="display:none;" type="submit" value="' +
-					 t('core', 'Send') + '" />';
-					 html += '</form>';
-					 }*/
+
+					var mailPublicNotificationEnabled = $('input:hidden[name=mailPublicNotificationEnabled]').val();
+					if (mailPublicNotificationEnabled === 'yes') {
+						html += '<form id="emailPrivateLink">';
+						html +=
+							'<input id="email" style="display:none; width:62%;" value="" placeholder="' +
+							t('core', 'Email link to person') + '" type="text" />';
+						html +=
+							'<input id="emailButton" style="display:none;" type="submit" value="' +
+							t('core', 'Send') + '" />';
+						html += '</form>';
+					}
 				}
 
 				html += '<div id="expiration">';
@@ -443,32 +443,32 @@
 						.append(insert)
 						.appendTo(ul);
 				};
-				// FIXME Emailing links is not supported in Gallery
-				/*if (link && linksAllowed && $('#email').length != 0) {
-				 $('#email').autocomplete({
-				 minLength: 1,
-				 source: function (search, response) {
-				 $.get(OC.filePath('core', 'ajax', 'share.php'), {
-				 fetch: 'getShareWithEmail',
-				 search: search.term
-				 }, function (result) {
-				 if (result.status == 'success' && result.data.length > 0) {
-				 response(result.data);
-				 }
-				 });
-				 },
-				 select: function (event, item) {
-				 $('#email').val(item.item.email);
-				 return false;
-				 }
-				 })
-				 .data("ui-autocomplete")._renderItem = function (ul, item) {
-				 return $('<li>')
-				 .append('<a>' + escapeHTML(item.displayname) + "<br>" +
-				 escapeHTML(item.email) + '</a>')
-				 .appendTo(ul);
-				 };
-				 }*/
+
+				if (link && linksAllowed && $('#email').length != 0) {
+					$('#email').autocomplete({
+						minLength: 1,
+						source: function (search, response) {
+							$.get(OC.filePath('core', 'ajax', 'share.php'), {
+								fetch: 'getShareWithEmail',
+								search: search.term
+							}, function (result) {
+								if (result.status == 'success' && result.data.length > 0) {
+									response(result.data);
+								}
+							});
+						},
+						select: function (event, item) {
+							$('#email').val(item.item.email);
+							return false;
+						}
+					})
+						.data("ui-autocomplete")._renderItem = function (ul, item) {
+						return $('<li>')
+							.append('<a>' + escapeHTML(item.displayname) + "<br>" +
+							escapeHTML(item.email) + '</a>')
+							.appendTo(ul);
+					};
+				}
 
 			} else {
 				html += '<input id="shareWith" type="text" placeholder="' +
@@ -721,15 +721,15 @@
 			if (mailNotificationEnabled === 'yes' &&
 				shareType !== this.SHARE_TYPE_REMOTE) {
 				var checked = '';
-				if (mailSend === '1') {
+				if (mailSend === 1) {
 					checked = 'checked';
 				}
 				html +=
-					'<input type="checkbox" class="checkbox checkbox--right" ' +
-					'name="mailNotification" class="mailNotification" ' +
+					'<input id="mail-' + escapeHTML(shareWith) + '" type="checkbox" class="mailNotification checkbox checkbox--right" ' +
+					'name="mailNotification" ' +
 					checked + ' />';
 				html +=
-					'<label>' + t('core', 'notify by email') + '</label>';
+					'<label for="mail-' + escapeHTML(shareWith) + '">' + t('core', 'notify by email') + '</label>';
 			}
 			if (oc_appconfig.core.resharingAllowed &&
 				(possiblePermissions & OC.PERMISSION_SHARE)) {
@@ -1171,73 +1171,69 @@ $(document).ready(function () {
 	});
 
 
-	// FIXME Emailing links is not supported in Gallery
-	/*$(document).on('submit', '#dropdown #emailPrivateLink', function (event) {
-	 event.preventDefault();
-	 var link = $('#linkText').val();
-	 var itemType = $('#dropdown').data('item-type');
-	 var itemSource = $('#dropdown').data('item-source');
-	 var file = $('tr').filterAttr('data-id', String(itemSource)).data('file');
-	 var email = $('#email').val();
-	 var expirationDate = '';
-	 if ($('#expirationCheckbox').is(':checked') === true) {
-	 expirationDate = $("#expirationDate").val();
-	 }
-	 if (email != '') {
-	 $('#email').prop('disabled', true);
-	 $('#email').val(t('core', 'Sending ...'));
-	 $('#emailButton').prop('disabled', true);
+	$(document).on('submit', '#dropdown #emailPrivateLink', function (event) {
+		event.preventDefault();
+		var link = $('#linkText').val();
+		var itemType = $('#dropdown').data('item-type');
+		var itemSource = $('#dropdown').data('item-source');
+		var fileName = $('.last').children()[0].innerText;
+		var email = $('#email').val();
+		var expirationDate = '';
+		if ($('#expirationCheckbox').is(':checked') === true) {
+			expirationDate = $("#expirationDate").val();
+		}
+		if (email != '') {
+			$('#email').prop('disabled', true);
+			$('#email').val(t('core', 'Sending ...'));
+			$('#emailButton').prop('disabled', true);
 
-	 $.post(OC.filePath('core', 'ajax', 'share.php'), {
-	 action: 'email',
-	 toaddress: email,
-	 link: link,
-	 itemType: itemType,
-	 itemSource: itemSource,
-	 file: file,
-	 expiration: expirationDate
-	 },
-	 function (result) {
-	 $('#email').prop('disabled', false);
-	 $('#emailButton').prop('disabled', false);
-	 if (result && result.status == 'success') {
-	 $('#email').css('font-weight', 'bold').val(t('core', 'Email sent'));
-	 setTimeout(function () {
-	 $('#email').css('font-weight', 'normal').val('');
-	 }, 2000);
-	 } else {
-	 OC.dialogs.alert(result.data.message, t('core', 'Error while sharing'));
-	 }
-	 });
-	 }
-	 });*/
+			$.post(OC.filePath('core', 'ajax', 'share.php'), {
+					action: 'email',
+					toaddress: email,
+					link: link,
+					file: fileName,
+					itemType: itemType,
+					itemSource: itemSource,
+					expiration: expirationDate
+				},
+				function (result) {
+					$('#email').prop('disabled', false);
+					$('#emailButton').prop('disabled', false);
+					if (result && result.status == 'success') {
+						$('#email').css('font-weight', 'bold').val(t('core', 'Email sent'));
+						setTimeout(function () {
+							$('#email').css('font-weight', 'normal').val('');
+						}, 2000);
+					} else {
+						OC.dialogs.alert(result.data.message, t('core', 'Error while sharing'));
+					}
+				});
+		}
+	});
 
-	// FIXME Emailing links is not supported in Gallery
-	/*$(document).on('click', '#dropdown input[name=mailNotification]', function () {
-	 var $li = $(this).closest('li');
-	 var itemType = $('#dropdown').data('item-type');
-	 var itemSource = $('#dropdown').data('item-source');
-	 var action = '';
-	 if (this.checked) {
-	 action = 'informRecipients';
-	 } else {
-	 action = 'informRecipientsDisabled';
-	 }
+	$(document).on('click', '#dropdown input[name=mailNotification]', function () {
+		var $li = $(this).closest('li');
+		var itemType = $('#dropdown').data('item-type');
+		var itemSource = $('a.share').data('item-source');
+		var action = '';
+		if (this.checked) {
+			action = 'informRecipients';
+		} else {
+			action = 'informRecipientsDisabled';
+		}
+		var shareType = $li.data('share-type');
+		var shareWith = $li.attr('data-share-with');
+		$.post(OC.filePath('core', 'ajax', 'share.php'), {
+			action: action,
+			recipient: shareWith,
+			shareType: shareType,
+			itemSource: itemSource,
+			itemType: itemType
+		}, function (result) {
+			if (result.status !== 'success') {
+				OC.dialogs.alert(t('core', result.data.message), t('core', 'Warning'));
+			}
+		});
 
-	 var shareType = $li.data('share-type');
-	 var shareWith = $li.attr('data-share-with');
-
-	 $.post(OC.filePath('core', 'ajax', 'share.php'), {
-	 action: action,
-	 recipient: shareWith,
-	 shareType: shareType,
-	 itemSource: itemSource,
-	 itemType: itemType
-	 }, function (result) {
-	 if (result.status !== 'success') {
-	 OC.dialogs.alert(t('core', result.data.message), t('core', 'Warning'));
-	 }
-	 });
-
-	 });*/
+	});
 });
--- a/templates/part.content.php
+++ b/templates/part.content.php
@@ -131,6 +131,10 @@ style(
 <div id="gallery" data-allow-public-upload="<?php p($_['publicUploadEnabled'])?>" class="hascontrols"></div>
 <div id="emptycontent" class="hidden"></div>
 <input type="hidden" name="allowShareWithLink" id="allowShareWithLink" value="yes"/>
+<input type="hidden" name="mailNotificationEnabled" id="mailNotificationEnabled"
+	   value="<?php p($_['mailNotificationEnabled']) ?>"/>
+<input type="hidden" name="mailPublicNotificationEnabled" id="mailPublicNotificationEnabled"
+	   value="<?php p($_['mailPublicNotificationEnabled']) ?>"/>
 <div class="hiddenuploadfield">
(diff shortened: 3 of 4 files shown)


controller/pagecontroller.php
```diff
@@ -80,9 +80,17 @@ public function __construct(
 	 */
 	public function index() {
 		$appName = $this->appName;
+		$config = \OC::$server->getConfig();
```
oparoz: IConfig is already injected and if it's not the right one, then you need to inject it from application.php instead of using `\OC`.

js/vendor/owncloud/share.js
```diff
@@ -256,19 +256,18 @@
...
-					 html += '<form id="emailPrivateLink">';
-					 html +=
-					 '<input id="email" style="display:none; width:62%;" value="" placeholder="' +
-					 t('core', 'Email link to person') + '" type="text" />';
-					 html +=
-					 '<input id="emailButton" style="display:none;" type="submit" value="' +
-					 t('core', 'Send') + '" />';
-					 html += '</form>';
-					 }*/
+					 html += '</div>';*/
```
oparoz: This looks strange.

imjalpreet: Strange as in?

oparoz: You don't need that line

imjalpreet: Actually, I was also thinking of removing that line but as in this PR it was in the commented out code, I didn't remove it. I have removed it in my PR for #565, which I am about to make.

oparoz:
That was easier than anticipated because the endpoints for email don't nee to be changed..
- [x] Not shown when admin setting is disabled
- [x] Sending an email shows "email sent"
- [x] Recipient gets email
- [x] Link in email goes straight into Gallery  

Now we need the same thing for internal shares :)

imjalpreet:
@oparoz I have a question. How can I get the Id of the album that is about to be shared? (like in this case: https://github.com/owncloud/core/blob/master/apps/files_sharing/js/sharetabview.js#L54)
Thanks!

Edit: No problems. I got it. :+1:

oparoz:
OK, but I'm introducing changes which will make it easier to get information about the current album.

imjalpreet:
Cool! That would ease it a lot.

imjalpreet:
@oparoz I have updated the PR with the option to send a mail to an internal user also. 
But I need your help, right now, it works but it checkbox resets everytime we click on dropdown(share button). I have found out that the reason for this is that in the ajax call to share.php we are sending wrong itemSource. It should be the albumId but we are sending the album path. I have been trying from last few hours but haven't been successful.
I had tried to get the albumId from gallery.js and galleryalbum.js but I am not able to get the correct id. By correct Id, I mean that at the time of displaying the album, we are getting other data attributes such as path, dir and some more, I tried to get Id also but what happens is that, we are creating the subAlbums when we are in a particular album, so I actually need the Id of the sub album but I am only able to get the Id of the current album. I hope you understood the issue.

oparoz:
> It should be the albumId but we are sending the album path.

If I'm right, I think you're confusing albumId and shareId. I haven't looked at the code, but I'll imagine it wants to know which share to send the email about.
If that's the case, that Id is available. Look how I retrieve it in other areas of the code.
If I'm wrong, I'll need to take a look.

> so I actually need the Id of the sub album but I am only able to get the Id of the current album

I don't think that's correct. The share button is about sharing the current album. To create a share we need the path, but to make modifications, we need the shareId

imjalpreet:
@oparoz, 

> If I'm right, I think you're confusing albumId and shareId. I haven't looked at the code, but I'll imagine it wants to know which share to send the email about.

No, as far as I have learnt from the code, it is the albumId we need to send in the ajax call to share.php (this is the case with the files app). It is because we need to include the name of the album in the email and marking the checkbox too. (The database is getting updated over [here](https://github.com/owncloud/core/blob/master/lib/private/share/share.php#L1079). Here the itemSource should be the albumId and not the path.

> I don't think that's correct. The share button is about sharing the current album. To create a share we need the path, but to make modifications, we need the shareId

I understand that the share button is for sharing the current album. I think you misunderstood me. What I meant was that when we displaying the albums and the images in the gallery and getting the path and other data values and assigning them to the data elements, I would include the albumId also over here. ([Here](https://github.com/owncloud/gallery/blob/master/js/galleryalbum.js#L5)) and [Here](https://github.com/owncloud/gallery/blob/master/js/galleryalbum.js#L91)

But this Album object is being created in the file gallery.js ([Here](https://github.com/owncloud/gallery/blob/master/js/gallery.js#L381) and [Here](https://github.com/owncloud/gallery/blob/master/js/gallery.js#L409)). The second is for the sub-albums  but when that function is called the `albumInfo.fileid` is still the main album but I need the fileid of the sub-album while displaying it. 

Secondly, I have tried passing the albumId in the ajax call to share.php(by hard-coding it) and it works perfectly but I am not able to find out a way how to do it the right way. 

This was my understanding. Correct me if I am wrong.
Thanks!

oparoz:
> What I meant was that when we displaying the albums and the images in the gallery and getting the path and other data values and assigning them to the data elements, I would include the albumId also over here. (Here) and Here

That's coming, but why do you need it for this PR? Those are for the sub-albums which can't be shared atm

js/vendor/owncloud/share.js
```diff
@@ -1174,73 +1173,68 @@ $(document).ready(function () {
...
-	 action = 'informRecipients';
-	 } else {
-	 action = 'informRecipientsDisabled';
-	 }
+	$(document).on('submit', '#dropdown #emailPrivateLink', function (event) {
+		event.preventDefault();
+		var link = $('#linkText').val();
+		var itemType = $('#dropdown').data('item-type');
+		var itemSource = $('#dropdown').data('item-source');
+		var file = $('tr').filterAttr('data-id', String(itemSource)).data('file');
```
oparoz: This doesn't exist in Gallery.

imjalpreet: Yeah, Removed. Sorry about that.

imjalpreet:
@oparoz, 

> That's coming, but why do you need it for this PR? Those are for the sub-albums which can't be shared atm

I will try to explain it again. I want the following:
1. According to me, we need to send the AlbumId as the itemSource variable in the POST request.
2. So, for getting the AlbumId I wanted to store it as a data attribute in the album thumbnail, so that I can use that later.
3. Secondly, I don't think those sub-albums cannot be shared because I can share them but I was just not able to figure out how to get their albumId.

**But these are not to be worried now, as I was writing this reply, I figured out a way to fix this issue. :smile: 
I am updating the PR, you can have a look.**
I hope everything is working now. 
Thanks! :+1:

imjalpreet:
@oparoz I would like to add to my previous reply. I just saw that the link in the email to the album shared, via internal share, is broken. This problem is not due to this PR and the same problem exists in the files view too. This problem is due to a mistake in [`/lib/private/share/mailnotifications.php`](https://github.com/owncloud/core/blob/master/lib/private/share/mailnotifications.php#L134)
So, changes are to be done in the above mentioned file, maybe one or two small changes would make it to work. 
Can you please look at it and give some feedback whether I am right or wrong?
Thanks!

oparoz:
> According to me, we need to send the AlbumId as the itemSource variable in the POST request.

Correct

> So, for getting the AlbumId I wanted to store it as a data attribute in the album thumbnail, so that I can use that later.

That's where I don't follow you. The current album is not shown in the photowall. It has no thumbnail.

> Secondly, I don't think those sub-albums cannot be shared because I can share them but I was just not able to figure out how to get their albumId.

All albums can be shared if you know their fileid, but there is no share button on an album

imjalpreet:
> That's where I don't follow you. The current album is not shown in the photowall. It has no thumbnail.

Though this is not required now as I found another way to get the albumId but I would clarify myself,
By Album Thumbnail, I meant the thumbnail of the subalbums that are there in the current album and by albumId I mean the Id of the particular sub-album. I hope you get me this time.

> All albums can be shared if you know their fileid, but there is no share button on an album

Yeah. I wanted to say the same thing that we need the albumId for sharing it when we are inside that album.

@oparoz Can you look at the modified code, in that both types of emails are working.

js/vendor/owncloud/share.js
```diff
@@ -145,7 +145,7 @@
 			var dropDownEl;
 			var self = this;
 			var html = '<div id="dropdown" class="drop shareDropDown" data-item-type="' + itemType +
-				'" data-item-source="' + path + '">';
+				'" data-item-source="' + path + '" data-id="' + data[0].file_source + '">';
```
oparoz: What happens when you open an album which has never been shared? I think you won't be able to send an email.

oparoz:
> By Album Thumbnail, I meant the thumbnail of the subalbums that are there in the current album and by albumId I mean the Id of the particular sub-album. 

OK, but why would you want that at this stage?

> Yeah. I wanted to say the same thing that we need the albumId for sharing it when we are inside that album.

albuminfo for the current album is sent every time we enter an album. It doesn't come from sub-albums. Maybe that's where the confusion is coming from.
The way the mapping is done in Gallery is unusual. Look at _mapFiles in gallery.js.

I think the best thing to do at this stage is to wait for my PR which sends more album information. That way we can store the albumid in data-item-source

js/vendor/owncloud/share.js
imjalpreet: Yeah, I forgot that test-case.

imjalpreet:
> I think the best thing to do at this stage is to wait for my PR which sends more album information. That way we can store the albumid in data-item-source

Ok. Cool. We have to store both the path as well as the albumid(maybe data-path and data-item-source). As we require both of them, one in the case of link-share and the other in case of internal share.

imjalpreet:
> albuminfo for the current album is sent every time we enter an album. It doesn't come from sub-albums. Maybe that's where the confusion is coming from.

Yeah. I had figured that out. That was the reason I got stuck there. Maybe, this wasn't the direction to go in at the very beginning.

js/vendor/owncloud/share.js
imjalpreet: This won't work when the dropdown is opened for the first share.

imjalpreet: @oparoz Do you find any other mistake in the rest of the code? This will be fixed, once your PR arrives.

imjalpreet:
@oparoz I have opened an issue regarding wrong link in email. [Here](https://github.com/owncloud/core/issues/23197)

oparoz:
Thanks.

oparoz:
`item-source` is now available https://github.com/owncloud/gallery/pull/593

imjalpreet:
Cool! Now the job is almost done. Just have to wait for it to get merged.  Thanks! :+1:

imjalpreet:
@oparoz I have made the required changes and tested it. I think it is completed.
Please give your feedback. Thanks!

oparoz:
OK, I will, but a couple of things.
- When you're working on your PR and still need guidance and feedback, set it to "developing" and ask me questions, etc. 
- when you think your PR is ready to review, please set it to "to review" and ask as many testers as you can to help you test the changes

It makes it easier to quickly know what needs reviewing, etc.

oparoz:
Argh and now it needs another rebase because I merged your other PR...

imjalpreet:
> Argh and now it needs another rebase because I merged your other PR...

No problems. I will rebase it. :+1: 

> OK, I will, but a couple of things.
> - When you're working on your PR and still need guidance and feedback, set it to "developing" and ask me questions, etc. 
> - when you think your PR is ready to review, please set it to "to review" and ask as many testers as you can to help you test the changes
> 
> It makes it easier to quickly know what needs reviewing, etc.

Cool! :+1:

imjalpreet:
@oparoz Done!

oparoz:
@rullzer can test as he's interested in using the Gallery class in Documents :)

oparoz:
#### TODO
- [x] Rebase
- [x] Update tests

imjalpreet:
@oparoz I have a question. I am not able to pass the tests for this as I am not able to find a way to write them correctly. I am getting the following failure message: 

```
There was 1 failure:

---------
1) OCA\Gallery\Controller\PageControllerTest::testIndex
Expectation failed for method name is equal to <string:getAppValue> when invoked 1 time(s)
Parameter 1 for invocation OCP\IConfig::getAppValue('core', 'shareapi_allow_public_upload', 'yes') does not match expected value.
Failed asserting that two strings are equal.--- Expected
+++ Actual
@@ @@
-'shareapi_allow_public_notification'
+'shareapi_allow_public_upload'
#1  /var/www/GSoC/apps/gallery/controller/pagecontroller.php:88
#2  /var/www/GSoC/apps/gallery/tests/unit/controller/PageControllerTest.php:90
#3  OCA\Gallery\Controller\PageControllerTest->testIndex

```

I am guessing we need to do something different if getAppValue is called multiple times but I am not able to figure out how.

oparoz:
> I am guessing we need to do something different if getAppValue is called multiple times but I am not able to figure out how

That's correct and the reason I asked you to use a map when writing the test in #604: https://github.com/owncloud/gallery/pull/604#discussion_r56527604

Ping me if not clear

imjalpreet:
@oparoz I have updated the PR with the required changes. Please have a look.

oparoz:
Thanks, changes look good, yes. Did the tests pass locally?
I need to figure out why you can't use the Github token on Travis

imjalpreet:
@oparoz Yes the tests passed locally. 
I am also surprised why did it give an error while trying to use the github token.

oparoz:
https://github.com/owncloud/gallery/pull/609

oparoz:
I've just merged it, please rebase and push. Thanks :)

imjalpreet:
Ohk. 
@oparoz Should I wait for it to get merged? I will have to send my PR again, right? (So that we can test it on Travis)

oparoz:
It's also a good idea to see what Scrutinizer has to say:

https://scrutinizer-ci.com/g/owncloud/gallery/inspections/39648f7c-fc5e-4171-8df2-e8f5ee39cfc6/issues/files/controller/pagecontroller.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

It helps you configure your IDE or maybe  should just pastebin my config

oparoz:
Seems to be working.

Still waiting for @rullzer to test this ;)

rullzer:
Ah missed this.. I'll test is soonish... added to the list :)

oparoz:
Thanks! It's just that you may want that patch for documents as well ;)

imjalpreet:
> It helps you configure your IDE or maybe should just pastebin my config

Sure. I will look at it too and configure my IDE. :+1:

oparoz:
The IDE config was published on the wiki.

controller/pagecontroller.php
```diff
@@ -85,7 +85,9 @@ public function index() {
 		$params = [
 			'appName' => $appName,
 			'uploadUrl' => $this->urlGenerator->linkTo('files', 'ajax/upload.php'),
-			'publicUploadEnabled' => $this->appConfig->getAppValue('core', 'shareapi_allow_public_upload', 'yes')
+			'publicUploadEnabled' => $this->appConfig->getAppValue('core', 'shareapi_allow_public_upload', 'yes'),
```
oparoz: Lines are too long. Max 100 characters

imjalpreet:
> The IDE config was published on the wiki.

Yeah. I have replaced my config file with that.

controller/pagecontroller.php
imjalpreet: @oparoz Sorry, forgot to change this. I had corrected it on my local machine but forgot to update the PR

oparoz:
Cool, I hope it works as-is :).
You just select your code block and do `ctrl+alt+l` to format your changes.
If you don't select anything, then the whole file gets formatted. That should be avoided because it makes it harder to review.
I usually do some clean-up PRs for files which need it.

imjalpreet:
@oparoz Updated it. :+1:

controller/pagecontroller.php
oparoz: The 1st one is still too long ;)

I'm just mentioning it because that was part of your previous PR

tests/unit/controller/PageControllerTest.php
```diff
@@ -72,12 +72,16 @@ public function testIndex() {
...
 		$params = [
 			'appName' => $this->appName,
 			'uploadUrl' => $url,
-			'publicUploadEnabled' => $publicUploadEnabled
+			'publicUploadEnabled' => $publicUploadEnabled,
+			'mailNotificationEnabled' => $mailNotificationEnabled,
+			'mailPublicNotificationEnabled' => $mailPublicNotificationEnabled
 		];
-		$this->mockGetPublicUploadAppValue($publicUploadEnabled);
+		$this->mockGetTestAppValue($publicUploadEnabled, $mailNotificationEnabled, $mailPublicNotificationEnabled);
```
oparoz: This might also be too long?

tests/unit/controller/PageControllerTest.php
```diff
@@ -235,15 +239,17 @@ private function mockUrlToDownloadPage($token, $fileId, $filename, $url) {
 						   ->willReturn($url);
 	}
 
-	private function mockGetPublicUploadAppValue($publicUploadEnabled) {
-		$this->appConfig->expects($this->once())
+	private function mockGetTestAppValue($publicUploadEnabled, $mailNotificationEnabled, $mailPublicNotificationEnabled) {
```
oparoz: And this one?

controller/pagecontroller.php
```diff
@@ -85,7 +85,13 @@ public function index() {
 		$params = [
 			'appName' => $appName,
 			'uploadUrl' => $this->urlGenerator->linkTo('files', 'ajax/upload.php'),
```
oparoz: I think this one is also too long and it's OK to shorten it in this PR

oparoz:
I hope you don't mind. All this nitpicking might seem useless, but it really helps keep code readable, especially in the browser.

It's also a good way to show you some side effects. Static analysers are pretty stupid and count real lines as opposed to lines of code and when you go above a certain number of lines, they complain and downgrade the method... which means that you need to refactor it slightly.
In this case, this would mean putting the array in `index()` in a separate private method.
You can do it here as an exercise or I'll do it later. As you want.

templates/part.content.php
```diff
@@ -131,6 +131,11 @@
 <div id="gallery" data-allow-public-upload="<?php p($_['publicUploadEnabled'])?>" class="hascontrols"></div>
 <div id="emptycontent" class="hidden"></div>
 <input type="hidden" name="allowShareWithLink" id="allowShareWithLink" value="yes"/>
+
```
oparoz: We don't need the extra space

templates/part.content.php
```diff
@@ -131,6 +131,11 @@
...
+<input type="hidden" name="mailNotificationEnabled" id="mailNotificationEnabled" value="<?php p($_['mailNotificationEnabled']) ?>" />
+
```
oparoz: Ditto

templates/part.content.php
```diff
@@ -131,6 +131,11 @@
 <div id="gallery" data-allow-public-upload="<?php p($_['publicUploadEnabled'])?>" class="hascontrols"></div>
 <div id="emptycontent" class="hidden"></div>
 <input type="hidden" name="allowShareWithLink" id="allowShareWithLink" value="yes"/>
+
+<input type="hidden" name="mailNotificationEnabled" id="mailNotificationEnabled" value="<?php p($_['mailNotificationEnabled']) ?>" />
```
oparoz: Probably too long

templates/part.content.php
```diff
@@ -131,6 +131,11 @@
...
+
+<input type="hidden" name="mailPublicNotificationEnabled" id="mailPublicNotificationEnabled" value="<?php p($_['mailPublicNotificationEnabled']) ?>" />
```
oparoz: Too long as well

controller/pagecontroller.php
imjalpreet: @oparoz What length should I try to maintain in all the lines?

imjalpreet:
@oparoz 

> I hope you don't mind. All this nitpicking might seem useless, but it really helps keep code readable, especially in the browser.

Yeah, No problems. It is a learning phase for me. So, I like the fact that you are trying to correct me.

> It's also a good way to show you some side effects. Static analysers are pretty stupid and count real lines as opposed to lines of code and when you go above a certain number of lines, they complain and downgrade the method... which means that you need to refactor it slightly.
> In this case, this would mean putting the array in index() in a separate private method.
> You can do it here as an exercise or I'll do it later. As you want.

In this case, you are talking about the params array, right? What exactly do you want me to do?

controller/pagecontroller.php
oparoz: @imjalpreet - 100 characters max, otherwise I get reports like these:
https://scrutinizer-ci.com/g/interfasys/galleryplus/inspections/b19f7e2a-00e4-42b3-b6e9-7aabc3e7899e/code-structure/operation/OCA%5CGalleryPlus%5CController%5CPageController%3A%3Aindex

oparoz:
> In this case, you are talking about the params array, right? What exactly do you want me to do?

Correct. Look at this commit: https://github.com/owncloud/gallery/commit/2b07fa3072c7aa4c31c1f8722474570630b9a389. I had to do the exact same thing because `getNodeData($node)` was flagged as being too long.

imjalpreet:
@oparoz Updated it! :+1:

controller/pagecontroller.php
```diff
@@ -82,11 +82,7 @@ public function index() {
 		$appName = $this->appName;
 
 		// Parameters sent to the template
-		$params = [
-			'appName' => $appName,
-			'uploadUrl' => $this->urlGenerator->linkTo('files', 'ajax/upload.php'),
-			'publicUploadEnabled' => $this->appConfig->getAppValue('core', 'shareapi_allow_public_upload', 'yes')
-		];
+		$params = $this->getParameters();
```
oparoz: Send `$appName`

controller/pagecontroller.php
```diff
@@ -96,6 +92,34 @@ public function index() {
 	}
 
 	/**
+	 * Returns the parameters to be used in the index function
+	 *
+	 * @return array<string, string>
+	 */
+	private function getParameters() {
+		$appName = $this->appName;
```
oparoz: Remove and use the received parameter.

oparoz:
Almost! :)

I also forgot to mention that the maximum number of lines allowed is 20, so it's really low and forces us to write short methods with a specific goal.

I found 2 problems when trying to keep methods shorts:
1. Finding meaningful names for all methods
2. Some classes are then flagged as being too complex because they have too many methods and they need to be refactored

imjalpreet:
@oparoz Done!

> Some classes are then flagged as being too complex because they have too many methods and they need to be refactored

What is done in this case?

controller/pagecontroller.php
```diff
@@ -96,6 +92,35 @@ public function index() {
 	}
 
 	/**
+	 * Returns the parameters to be used in the index function
+	 *
+	 * @param $appName
+	 *
+	 * @return array<string, string>
```
oparoz: Typo, no space allowed after the comma

controller/pagecontroller.php
```diff
@@ -96,6 +92,35 @@ public function index() {
...
+	 */
+	private function getParameters($appName) {
```
oparoz: I think we should be more specific, like `getIndexParameters`. The reason being that almost all methods on this page have parameters, so if we need to do this another time this would have to be changed.

oparoz: Also, this is a private method. The order in PHP classes for this project is

public
protected
private

So this method should be moved down as the first private method, before `showPublicPage`

oparoz:
> What is done in this case?

You split the class in two usually. When looking at it, you always find a group of methods which are trying to solve a specific problem and which should probably be in their own class. But that's just one solution. Sometimes you sub-class, etc.

oparoz:
And sorry for finding all these little things now. I didn't think to check for those earlier.

imjalpreet:
@oparoz Done!

> And sorry for finding all these little things now. I didn't think to check for those earlier.

I have no issues. As I told you, I am learning from all this.

oparoz:
Thanks for your PR and good job :)

https://scrutinizer-ci.com/g/owncloud/gallery/inspections/e9a316fe-ba9c-4c06-a6b0-d0631d3ea7cf

tests/unit/controller/PageControllerTest.php
```diff
@@ -72,12 +72,18 @@ public function testIndex() {
...
 		$publicUploadEnabled = 'yes';
+		$mailNotificationEnabled = 'no';
+		$mailPublicNotificationEnabled = 'yes';
 		$params = [
 			'appName' => $this->appName,
 			'uploadUrl' => $url,
-			'publicUploadEnabled' => $publicUploadEnabled
+			'publicUploadEnabled' => $publicUploadEnabled,
+			'mailNotificationEnabled' => $mailNotificationEnabled,
+			'mailPublicNotificationEnabled' => $mailPublicNotificationEnabled
```
oparoz, line 82: One thing I've just noticed, for next time, the `=>` are not aligned.

This is what I mean:

```
        $params = [
            'appName'                       => $this->appName,
            'uploadUrl'                     => $url,
            'publicUploadEnabled'           => $publicUploadEnabled,
            'mailNotificationEnabled'       => $mailNotificationEnabled,
            'mailPublicNotificationEnabled' => $mailPublicNotificationEnabled
        ];
```

imjalpreet, line 82: Noted. :+1:

Milo-W:
Doesn't work! No sharing button!

Issue #224, opened 2015-07-22
Demo content


oparoz:
When I put together [the demo site](https://oc8demo.interfacloud.com/index.php/apps/galleryplus/s/4jjwPv5E21l8Lnk), it was to showcase the new formats which Gallery could show previews of, but I think we could also have another demo with lots of content, so that we can test and show how the app behaves in a situation which is closer to how people are using ownCloud.

So if you have content you're willing to "showcase", please let me know.

It's probably not possible to import folders from external locations, so I would need to re-upload the content unless someone has a better idea. Note that the content is encrypted, so afaik I can't just fetch it from the server and move it to the data folder.

I would also need a little note about the collection or at least a copyright statement.

@jospoortvliet @jancborchardt @demattin @setnes @libasys 
##

jancborchardt:
For me it’s especially important that the Gallery app nicely handles:
- SVG files
- small icons, not upscaling them in an ugly fashion but still making them visible
- images with transparency, be it PNG or SVG – especially black ones. Wikipedia handles this very nicely _without_ a switcher: https://github.com/owncloud/galleryplus/issues/226

Test files for black transparent PNG https://github.com/opensourcedesign/opensourcedesign.github.io/blob/master/images/foss-logo-512.png and SVG https://github.com/opensourcedesign/opensourcedesign.github.io/blob/master/images/foss-logo-512.svg

oparoz:
SVG files are disabled by default for security reasons, but if enabled, browsers do an OK job, as long as there isn't too much transparency.

Small icons are a tough one. You can't have a photowall made of 50x50 GIFs, so there has to be some scaling, but the admin decides how much.

oparoz:
@jancborchardt I think your list should be converted into acceptance tests so that we can work on each individual case if it doesn't match your expectations.

We should still try to come up with as much content as possible and offer a download so that people can use that to quickly test PRs, etc.

oparoz:
This issue was moved to nextcloud/gallery#53

PR #594, opened 2016-03-14
Update README.md

aspdye:
- Better Screenshot (provided by screenshots-repository)
- Changed "Gallery" to "ownCloud Gallery" (like in all the other repositories like Mail, Calendar, Contacts and Core)
- Changed some small appearance things
- [ ] The Gallery+ part (This version is for ownCloud 9.1) is a bit confusing :confused: 

@oparoz @DeepDiver1975

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,14 @@
-# Gallery 
+# ownCloud Gallery 
 [![Build Status](https://travis-ci.org/owncloud/gallery.svg?branch=master)](https://travis-ci.org/owncloud/gallery)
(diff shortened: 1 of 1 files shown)


oparoz:
Thanks for taking care of that. I agree we need a new screenshot and I used what I had available because the screenshot in the repository you mentioned is not showing enough of what Gallery can do.

### Screenshot requirements
- Default theme
- Albums
- Info button
- Share button

aspdye:
You are right. Maybe you can select another one of those: https://github.com/owncloud/screenshots/tree/master/gallery

If this still does not meet the requirements - then we should make a new screenshot and add it into the screenshots-repository :wink:

oparoz:
Half of the screenshots there are wrong :D 

I did make a [call for content](https://github.com/owncloud/gallery/issues/224) to help make a better demo site, but nothing has changed since July last year.

PR #595, opened 2016-03-14
Only show sort buttons if there is more than 1 image in the album

virajparimi:
Fixes: #573 

Licence: AGPL
### Description

Sorted buttons only show up when there is more than one image or folder in the gallery.
### Features

Added  checks to ensure that no typeError problems occur due to usage of `album.images` or `album.subAlbums`.
### Tested on
- [X] Ubuntu 14.04/Chrome
### TODO

Inorder to test,
- [1]  Add a .nomedia file to folder which has a parent folder.
- [2]  Switch to Gallery.
- [3]  Press on the parent folder on the breadcrumb.

Then, ensure no typeError problem is coming.
### Reviewers

@oparoz

### DIFF
--- a/js/galleryview.js
+++ b/js/galleryview.js
@@ -233,8 +233,19 @@
 			var availableWidth = $(window).width() - Gallery.buttonsWidth;
 			this.breadcrumb.init(albumPath, availableWidth);
 			var album = Gallery.albumMap[albumPath];
-			
-			var sum = album.images.length + album.subAlbums.length;
+			var sum = 0;
+			if(!_.isUndefined(album.images) && !_.isUndefined(album.subAlbums))
+			{
+				sum = album.images.length + album.subAlbums.length;
+			}
+			else if(_.isUndefined(album.images) && !_.isUndefined(album.subAlbums))
+			{
+				sum = album.subAlbums.length;	
+			}
+			else if(!_.isUndefined(album.images) && _.isUndefined(album.subAlbums))
+			{
+				sum = album.images.length;
+			}
 			//If sum of the number of images and subalbums exceeds 1 then show the buttons.
 			if(sum > 1)
 			{


virajparimi:
@oparoz Please have a look at this and give your feedback.

oparoz:
Thanks. I don't think you've tested your solution ;)
If you had, you would have encountered another bug. I think the problem is more serious and your recent change is just highlighting the problem introduced by: https://github.com/owncloud/gallery/commit/0e89aab141d46458cb714ad5467754fdbb418924

virajparimi:
I am a bit confused as I am still unable to reproduce the error mentioned by you, and yes I did click on the parent folder breadcrumb but still no error came. So I thought, to be on the safe side I added the appropriate checks. And checked if the original functionality was preserved or not. 

> If you had, you would have encountered another bug. I think the problem is more serious and your recent change is just highlighting the problem introduced by: 
> 0e89aab

Are you referring to the error/warning message I showed you previously?
Here are some screenshots,
First I tried to upload abc.nomedia file with 0 bytes.
onwcloud_2

Then I tried adding file with some information.
owncloud_6

I clicked on the  gallery app.
owncloud_4

Then clicked on the parent folder.
owncloud_5

Am I doing something wrong somewhere?

oparoz:
> I am a bit confused as I am still unable to reproduce the error mentioned by you, and yes I did click on the parent folder breadcrumb but still no error came

OK, that's because my original report was missing something. The parent folder must contain images.

> So I thought, to be on the safe side I added the appropriate checks. And checked if the original functionality was preserved or not.

The problem with that approach is that you can't test your own code

> Are you referring to the error/warning message I showed you previously?

No, I reworked the way decisions are made to show the empty message and apparently, the controls bar is initialised without having received enough information in some cases.

oparoz:
I've updated the instructions https://github.com/owncloud/gallery/issues/573#issuecomment-196018098

virajparimi:
Ok, now I have been able to reproduce the typeError that used to come, but the code I gave did eliminate that error. 

> If you had, you would have encountered another bug.

What other bug are you talking about?

oparoz:
> What other bug are you talking about?

`TypeError: Gallery.albumMap[Gallery.currentAlbum].images is undefined`

I haven't had time yet to investigate...

virajparimi:
So, I guess that where ever `album.images` or `album.sumAlbums` is used in the file gallerview.js, there we need to put appropriate checks.

oparoz:
Maybe... The logic needs to be checked. It's normal for the buttons to be initialised, even if the album is empty, so maybe only this method needs fixing.

oparoz:
Thanks for trying to fix this. The solution is in https://github.com/owncloud/gallery/pull/597

PR #597, opened 2016-03-15
Build the albumMap even when the server returns a 403

oparoz:
### Description

Album navigation breaks when the starting point is an empty Gallery and the user moves to the parent folder.
That's because the map we use for navigation is not built when the server returns an error status to a getFiles request.
#### Steps to reproduce
1. Add pictures to a folder
2. Create a sub-folder
3. Enter the sub-folder
4. Add a `.nomedia` file
5. Switch to Gallery
6. Press on the parent folder in the breadcrumb
#### Result before

```
TypeError: album.images is undefined  in galleryview.js
```
#### Result with this PR

No error
### Tested on
- [x] Windows/Firefox
- [x] Windows/Chrome
### Reviewers

@tahaalibra @viraj96 @imjalpreet @mbtamuli @rahulgoyal030

### DIFF
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -93,7 +93,9 @@
 
 			}, function (xhr) {
 				var result = xhr.responseJSON;
-				Gallery.view.init(decodeURIComponent(currentLocation), result.message);
+				var albumPath = decodeURIComponent(currentLocation);
+				Gallery.view.init(albumPath, result.message);
+				Gallery._mapStructure(albumPath);
 			});
 		},
 


virajparimi:
I have tested it, and it works fine now. No error encountered by me. :+1:

mbtamuli:
I too found no errors. :+1: Works correctly.

oparoz:
Thanks for testing the fix. If I could ask you to keep using this branch to do normal operation to see if there are no side-effects, I would appreciate it as I didn't have time to thoroughly test this:
- Switching between Files and Gallery always works
- No problem on the public side
- Uploading still works
- Drag and dropping still works
- etc.

The same effect will be achieved if I merge this, but a pre-run is always better ;)

oparoz:
Actually, even better would be for you guys to merge this PR with https://github.com/owncloud/gallery/pull/593 in your dev environment. Just cherry pick the 2 commits into a new fork of master and use that when playing with Gallery.
The sooner those get merged, the sooner we can add some new features :)

imjalpreet:
@oparoz I have tested it by merging this with #593 and I have tried to test a few things and they seem to work well :+1: 
- [x] The test to check this PR.
- [x] Public Side is working
- [x] Switching between Files and Gallery
- [x] Uploading Files/Folders
- [x] Uploading using drag and drop
- [x] Transfer files/folders using drag/drop
- [x] Both sorts checked

I will continue to test this and get back to you if I find any error.

oparoz:
Perfect, thank you

Issue #601, opened 2016-03-16
Slideshow/Image Preview not working in public share link.


imjalpreet:
### Steps to reproduce

As a logged in user,
1. Generate a public share link.
2. Go to that link.
3. Click on any image.
### Expected behaviour

BigShot container should show up, i.e., Image should show up in the BigShot container.
### Actual behaviour

Nothing happens and there is this error in the console: 
`Uncaught TypeError: Cannot read property 'hideDropDown' of undefined`
### Server configuration

**Operating system**: Ubuntu 14.04

**Web server:**Apache/2.4.7

**Database:** mysql Ver 14.14 Distrib 5.6.28, for debian-linux-gnu (x86_64)

**PHP version:** PHP 5.5.9-1ubuntu4.14

**ownCloud version:** 9.1.0 pre alpha

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - dav: 0.2.1
  - federatedfilesharing: 0.2.0
  - federation: 0.1.0
  - files: 1.5.1
  - files_sharing: 0.10.0
  - files_trashbin: 0.9.0
  - files_versions: 1.3.0
  - gallery: 15.0.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - updatenotification: 0.2.0
Disabled:
  - encryption
  - files_external
  - testing
  - user_ldap
```

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc4z46udfiog",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "localhost:8080"
        ],
        "datadirectory": "\/var\/www\/GSoC\/data",
        "overwrite.cli.url": "http:\/\/localhost:8080",
        "dbtype": "mysql",
        "version": "9.1.0.0",
        "dbname": "ownCloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
        "installed": true,
        "mail_from_address": "jalpreet.cse",
        "mail_smtpmode": "smtp",
        "mail_domain": "gmail.com",
        "mail_smtpsecure": "ssl",
        "mail_smtpauthtype": "LOGIN",
        "mail_smtpauth": 1,
[...]

PR #602, opened 2016-03-16
Close the share dialogue when performing another action, fixes #545

### LINKED ISSUE #545: Close the share dialogue when performing another action

virajparimi:
Fixes: #545 

Licence: AGPL
### Description

Initial commit had a regression. Has been fixed by adding `hideDropDown` button in the gallery.js file
### Features

Should fix the errors related to share dialog not closing.
### Test Plan

First click on the share button, then
- [ ] Click on any album.
- [ ] Click on any image.
- [ ] Click on infobutton.

Then,
- [ ] Similarly do the same on public side after sharing a file/folder.

Then,
- [ ] Similarly do the  same on the files app side too. 

Should not give any errors and share dialog should close.
### Tested on
- [X] Ubuntu 14.04/Chrome
### Reviewers

@oparoz @imjalpreet  @tahaalibra @mbtamuli

### DIFF
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -260,7 +260,6 @@
 			emptyContentElement.html(message);
 			emptyContentElement.removeClass('hidden');
 			$('#controls').addClass('hidden');
-			$('#loading-indicator').hide();
 		},
 
 		/**
--- a/js/gallery.js
+++ b/js/gallery.js
@@ -360,7 +360,10 @@
 				}
 			};
 			Gallery.activeSlideShow.show(start);
-
+			if(!_.isUndefined(Gallery.Share)){
+				Gallery.Share.hideDropDown();
+			}
+			$('.album-info-container').slideUp();
 			// Resets the last focused element
 			document.activeElement.blur();
 		},
--- a/css/public.css
+++ b/css/public.css
@@ -48,10 +48,3 @@ body {
 #gallery.hascontrols {
 	padding-bottom: 0;
 }
-
-#loading-indicator {
-	height: 32px;
-	width: 100%;
-	margin-top: 100px;
-	margin-bottom: 0;
-}
--- a/css/styles.css
+++ b/css/styles.css
@@ -100,7 +100,8 @@ div.crumb.last a {
 	vertical-align: top;
 }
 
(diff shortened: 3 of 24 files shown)


virajparimi:
@oparoz Sorry for the late reply...Have a look at it. Inspect element still gives the error but does not stop slideshow from being loaded from the public side.

oparoz:
1. The fix seems to be incomplete as the offending code hasn't been removed from slideshow.js
2. You need to fix the newly reported bug as well https://github.com/owncloud/gallery/issues/601

virajparimi:
@oparoz How does this look now?

js/gallery.js
```diff
@@ -360,7 +360,10 @@
 				}
 			};
 			Gallery.activeSlideShow.show(start);
-
+			if(Gallery.Share){
+				Gallery.Share.hideDropDown();
+				$('.album-info-container').slideUp();
```
oparoz: This call has nothing to do with Gallery.Share

virajparimi:
So, I need to check that `Gallery.Share` is undefined even in the other files where I called it like [here](https://github.com/owncloud/gallery/blob/master/js/galleryinfobox.js#L29) and [here](https://github.com/owncloud/gallery/blob/master/js/galleryalbum.js#L214) including this one.

oparoz:
Correct, but don't include the infoButton as that should still be available on the public view.

virajparimi:
@oparoz Updated the files as discussed. Please have  a look.

oparoz:
I think that should work.

Please include your test plan in the OP so that testers can review your changes

imjalpreet:
Tested it. Working fine. :+1:

oparoz:
Thanks @imjalpreet 

@viraj96 You forgot one case in your plan. The slideshow needs to work on the Files side.
Also, you need to squash your commits.

virajparimi:
@oparoz I am closing this as I am going to send another one. Sorry for the inconvenience.

PR #616, opened 2016-03-20
Fix ported from core 'Link share does not show all files' https://git…

alxerg:
Fixes: _Issue similar to 'Link share does not show all files' owncloud/core/issues/22894_ see: owncloud/core/pull/22548

### DIFF
--- a/css/public.css
+++ b/css/public.css
@@ -7,7 +7,6 @@ body {
 #content {
 	height: initial;
 	min-height: calc( 100vh - 120px );
-	overflow: hidden;
 }
 
 /* within #save */


oparoz:
Thanks for your contribution, but I don't think Gallery is affected by this bug. I've just shared 200 files and was able to scroll down until the end with both Files and Gallery.

If you are affected, you need to provide detailed steps on how to reproduce the issue.

alxerg:
Thanks for your attention :).
I'm using clean owncloud 8.2.2 and I've found this bug while viewing shared links in Chrome 49.0.2623.87 (64-bit) on Mac OS and Chrome on WinXP. The Gallery which shipped with owncloud 8.2.2 had no modifications. Provided fix works fine for me, perhaps it doesn't appear with more recent versions of the core and it's a kind of backporting thing, I'm not sure.

In more details, I had the issue for both Files and Gallery in Chrome on Mac OS. With 160 image files in shared folder the amount of displayed files was limited to the visible area of the screen and scrolling didn't work for both Files and Gallery. It was near 15 files visible in list mode ( Files ) and near 20 files in thumbnails mode ( Gallery ). Patching with https://github.com/owncloud/core/pull/22548 fixed the issue for Files but had to effect on the thumbnails mode ( Gallery ). It was still limited amount of visible thumbnails and no reaction on scroll there. After applying similar patch to https://github.com/owncloud/gallery/css/public.css all 160 thumbnails loaded fine.

oparoz:
I think the problem is that you have pdf_viewer installed. Uninstall it and you won't need any patches.

alxerg:
Understood, perhaps you are right. However some of my files are pdfs and pdf_viewer is quite handy to work with them, I'm going to keep using it. Ok, for now everything works quite fine for me, perhaps I'll be able to check this more detailed later. Thanks for your work for this helpful project!

oparoz:
I think you'll be fine once you update to 8.2.3. pdf_viewer was fixed in that release. It's easier to fix that one app than to patch all others ;)

PR #618, opened 2016-03-22
Don't ask to load new rows until the last operation is completed

### LINKED ISSUE #614: Scrolling when the photowall is still rendering will break rows
# Bug report
#### Setup
1. Create a test folder and a sub-folder from the files app
2. Fill the 1st folder with lots of images
3. Create a sub-folder
4. Add a few pictures to the folder, just enough to notice a difference if some images not belonging to this folder are added
5. Switch to Gallery
6. If you can't scroll down, add more images or resize your browser window
#### Test 1
1. Start from the top folder
2. Scroll down a little bit
3. Upload an image
#### Test 2
1. Start from the top folder
2. Refresh the page
3. Scroll down fast as soon as you can

_Note: It's more noticeable if the rows are 5+ images wide_
#### Test 3
1. Start from the sub-folder
2. Click on the breadcrumb to go to the parent folder
3. As soon as you see the album for the sub-folder, click on it
### Expected behaviour

The photowall should be rendered without any breaks or extra images
### Actual behaviour

- The photowall breaks, right where the image you've uploaded is located
- The photowall has some rows which are shorter than the width
- The photowall includes rows which belong to the parent album
### Server configuration

**ownCloud version:** 8.2 to 9.1

**Are you using external storage, if yes which one:** No

**Are you using encryption:** yes

**Are you using custom gallery.cnf config files:** yes
#### ownCloud log (data/owncloud.log)

Nothing relevant
### Client configuration

**Browser:** All

**Operating system:** All
#### Browser log

Nothing

oparoz:
Fixes: #614 
### Description

If you scroll down while the rows are rendering, some images can end up on the wrong rows and the photowall won't be seamless
Similarly, if you scroll down a bit and then decide to upload an image, the rendering starts and is disrupted after a few rows.

My solution is to change the way we detect that an operation is still ongoing.
### Caveats

I haven't noticed any difference in rendering speed on my setup, but let me know if something doesn't feel right.
### Test plan

The tests in #614 will show you what happens without this fix.
- [x] #614 test1
- [x] #614 test2
- [x] #614 test3 
- [x] Move images out of the current album
- [x] Go inside an album and come back very quickly
- [x] Resize the width or height of an album (There is still another bug hidden in there)

Basically, abuse Gallery's photowall and see if it breaks.
### Tested on
- [x] Windows/Firefox
- [x] Windows/Chrome
### Todo
- [x] Fix the case when users resize the window after having scrolled down
### Check list
- [x] Code is properly formatted
- [x] Commits have been squashed
- [x] Tests (unit, integration, api and/or acceptance) are included
### Reviewers

@manishbisht @tahaalibra @viraj96 @imjalpreet @mbtamuli @rahulgoyal030 @shubhrajodiya @harshjv @kartik69 @senorsen

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -49,8 +49,7 @@ $(document).ready(function () {
 
 		// This block loads new rows
 		$('html, #content-wrapper').scroll(function () {
-			Gallery.view.loadVisibleRows(Gallery.albumMap[Gallery.currentAlbum],
-				Gallery.currentAlbum);
+			Gallery.view.loadVisibleRows(Gallery.albumMap[Gallery.currentAlbum]);
 		});
 
 
--- a/js/galleryalbum.js
+++ b/js/galleryalbum.js
@@ -54,6 +54,7 @@
 	};
 
 	Album.prototype = {
+		requestId: null,
 		droppableOptions: {
 			accept: '#gallery > .row > a',
 			activeClass: 'album-droppable',
@@ -86,12 +87,11 @@
 		 * Returns a new album row
 		 *
 		 * @param {number} width
-		 * @param requestId
 		 *
 		 * @returns {Gallery.Row}
 		 */
-		getRow: function (width, requestId) {
-			return new Gallery.Row(width, requestId);
+		getRow: function (width) {
+			return new Gallery.Row(width);
 		},
 
 		/**
--- a/js/galleryrow.js
+++ b/js/galleryrow.js
@@ -5,14 +5,12 @@
 	 * Creates a row
 	 *
 	 * @param {number} targetWidth
-	 * @param {number} requestId
 	 * @constructor
 	 */
-	var Row = function (targetWidth, requestId) {
+	var Row = function (targetWidth) {
 		this.targetWidth = targetWidth;
 		this.items = [];
 		this.width = 4; // 4px margin to start with
-		this.requestId = requestId;
 		this.domDef = $('<div/>').addClass('row');
 	};
 
--- a/js/galleryview.js
+++ b/js/galleryview.js
@@ -29,6 +29,8 @@
 		 * Removes all thumbnails from the view
 		 */
 		clear: function () {
+			this.loadVisibleRows.processing = false;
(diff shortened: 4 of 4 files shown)


virajparimi:
When I resize the window I dont get any pictures in the photowall.
Steps I followed:-
1. Album is full of images
owncloud1
2. I resize the window and images are still there.
owncloud2
3. But if I resize it again after scrolling then no images are shown.
owncloud3
Are you referring to similar bug that I am facing?

oparoz:
> Are you referring to similar bug that I am facing?

Yes, that's the failing test in the OP. 
That change I did is not enough as it also blocks legitimate requests.

virajparimi:
If this is the only case where it is failing, then can we use `window.resize` event to detect the browser width change and then add another if statement to cover that case too.

oparoz:
@viraj96 - I've made a change. The state wasn't cleared when emptying the view. Let's hope this works in all situations...

oparoz:
@viraj96 It still doesn't work properly... The row loader doesn't load more rows once the user has reached the bottom of the window.

virajparimi:
Aren't the conditions in the if statement doing the same thing? If yes, why do a double check.

oparoz:
I supposed you're making a reference to `this.loadVisibleRows.loading`. It's a mixed variable, not only Boolean.
(But maybe it doesn't matter for this check. I don't trust Javascript, but this can be improved later)

oparoz:
@manishbisht @tahaalibra @viraj96 @imjalpreet @mbtamuli @rahulgoyal030 @shubhrajodiya @harshjv @kartik69 @senorsen

This needs testing

oparoz:
An easy way to test is to use [Gallery+](https://apps.owncloud.com/content/show.php/Gallery%2B?content=169116) which already includes this fix.

jospoortvliet:
Played with Gallery+, couldn't find any problems. Love the loading of the images, looks really great. :+1: from me.

oparoz:
Thank you for testing @jospoortvliet !

Issue #619, opened 2016-03-23
galleryfileaction.js causes redirect loop if used in unauthenticaed route


rullzer:
### Steps to reproduce
1. Have an app `files_foo` that has a public route.
2. Access that route
### Expected behaviour

All is fine
### Actual behaviour

Page gets reloaded over and over again.

This happens because of 
1. https://github.com/owncloud/gallery/blob/master/appinfo/app.php#L65
2. which loads https://github.com/owncloud/gallery/blob/master/appinfo/app.php#L75
3. which in turn does https://github.com/owncloud/gallery/blob/master/js/galleryfileaction.js#L214-L218
4. Which fails with a 401
5. which is then caught globally and reloads

Easies solution is probabaly to set `allowAuthErrors` https://github.com/owncloud/core/blob/v9.0.0/core/js/js.js#L1442

PVince81:
Why is "galleryfileaction.js" even returning a 401 ? A 404 or 500 I would understand.
Is the code loading the file missing authentication info / cookies ?

rullzer:
@PVince81 it is requesting: `../apps/gallery/config?extramediatypes=1` which is an authenticated route...

oparoz:
It's probably easier to solve if there is a standard way to retrieve a token to fix this line:
https://github.com/owncloud/gallery/blob/master/js/galleryfileaction.js#L22

PVince81:
Would this regexp https://github.com/owncloud/gallery/blob/master/appinfo/app.php#L65 also match other apps like "files_xxx" ?

PVince81:
This approach can be used now to only load scripts in the files app: https://github.com/owncloud/core/blob/v8.2.2/apps/files_sharing/appinfo/app.php#L57

oparoz:
> Would this regexp https://github.com/owncloud/gallery/blob/master/appinfo/app.php#L65 also match other apps like "files_xxx" ?

I'll check, but this may very well be the case

> This approach can be used now to only load scripts in the files app: https://github.com/owncloud/core/blob/v8.2.2/apps/files_sharing/appinfo/app.php#L57

It can't be used yet. See: https://github.com/owncloud/core/issues/22328

oparoz:
@PVince81 - The regexp doesn't match `files_xxx`

oparoz:
But if that app has a public, short URL using `/s/`, then yes, the script will be loaded.

rullzer:
[...]

PR #620, opened 2016-03-24
Fix unauthenticated route in files_drop app

rullzer:
Quick hack to fix the (enterprise) files_drop app (see #619 )

I know it is not ideal. but probabaly the best we can do with the time constrainsts...

CC: @oparoz @PVince81

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -62,7 +62,7 @@ Util::addTranslations('gallery');
 $request = $c->query('Request');
 if (isset($request->server['REQUEST_URI'])) {
 	$url = $request->server['REQUEST_URI'];
-	if (preg_match('%/apps/files(/.*)?%', $url)
+	if (preg_match('%/apps/files(_sharing)?/%', $url)
 		|| preg_match('%^((?!/apps/).)*/s/\b(.*)\b(?<!/authenticate)$%', $url)
 	) {
 		// @codeCoverageIgnoreStart


oparoz:
### To test
- [x] slideshow still works in Files
- [x] slideshow still works on public page
- [x] files_drop doesn't load any Gallery scripts

appinfo/app.php
```diff
@@ -62,8 +62,9 @@ function () use ($c, $appName) {
 $request = $c->query('Request');
 if (isset($request->server['REQUEST_URI'])) {
 	$url = $request->server['REQUEST_URI'];
-	if (preg_match('%/apps/files(/.*)?%', $url)
-		|| preg_match('%^((?!/apps/).)*/s/\b(.*)\b(?<!/authenticate)$%', $url)
+	if (!preg_match('%/apps/files_drop/%', $url) &&
```
PVince81: if the gallery app needs "files" and "files_sharing", how about adding these two in an opt-in regexp instead of doing an opt-out of files_drop ?

oparoz: Probably a good idea

rullzer: agreed. done

PVince81:
Looks good, slideshow also still works in regular file list and slideshow :+1:

oparoz:
Tested and works :+1: 
Thanks @rullzer

PVince81:
stable9: https://github.com/owncloud/gallery/pull/622

PVince81:
Reverted through https://github.com/owncloud/gallery/issues/629

Issue #547, opened 2016-03-01
Create .nomedia file from the album itself


oparoz:
## Feature request

**User type**: All

**User level**: All
### Description

Add an entry to the [+] menu allowing users to quickly blacklist an album from Gallery. The menu entry will create the `.nomedia` file

_Note: The [+] menu is only available from 9.1, after merging ~~#520~~ #574_
### Benefit / value

It just avoids having to do a round trip via Files just for that.
### Risk / caveats

None
##

prastut:
How can I blacklist an album from the files menu? When I click the [+] menu all I got is the feature to upload and create a new folder.

OC version:  9.1.0 pre alpha (git)

oparoz:
This has to happen from the Gallery side and can only be implemented once this is merged:
~~#520~~ #574.
That PR needs a fix to make it work with a last minute change in `core`. You're welcome to try and fix that, but that requires learning about how the uploading mechanism works in Files and what the equivalent variables might be in Gallery.

tahaalibra:
i want to try this fix....i guess now that #574 replaces #520...and it in the test it works great..i can now get at this issue

oparoz:
:+1:, but make sure you fork #574 as your base as it hasn't been merged yet.

oparoz:
This can now be worked on

oparoz:
Implemented in #625

PR #625, opened 2016-03-25
Create .nomedia file from the album itself

eknoes:
Licence: MIT

Adapted owncloud/core@d3861685045a006b47c68185223be21a6233802e to make work on  #547 easier.
### Description

This pull request makes it possible, to create a .nomedia file from the album itself via the [+] button. Furthermore it adapts owncloud/core@d3861685045a006b47c68185223be21a6233802e, so that there is a [standardized way](https://doc.owncloud.org/server/9.0/developer_manual/app/js.html#extending-the-new-menu-in-the-files-app) to extend the [+]-menu.
### Features
- Possibility to extend the [+] buttons menu
- Creation of a .nomedia file from the album itself
### Test plan
- [x] "Hide Album" Submenu is displayed
- [x] Clicking Hide-Album creates a .nomedia file
- [x] Gallery will not show up the album anymore
### Tested on
- [x] Linux/Firefox
- [x] Android 5.1/Chrome
### TODO
- [x] Implement extending [+] button menu
- [x] Document extending
- [x] Implement creation of .nomedia file
- [x] Implement what follows after hiding an album and you are "inside"
- [x] Test
### Check list
- [X] Code is properly formatted
- [ ] Commits have been squashed

This is my first pull request, so if you have any advice for me please feel welcome.


oparoz:
Thank you very much for this PR. I'll take a look shortly.
1st impression is good, Code looks properly formatted, etc.

It's usually a good idea to keep some of the relevant elements in the template, especially the checkboxes regarding documentation, etc. If they're ticked, it means you're aware of the requirements and have done what was asked of you.
And per example, how did you test that it works? Maybe try to fix #547 in a 2nd commit?

Also, unless you sign the contributor agreement, you should add to your OP that this is MIT licensed.

eknoes:
Fixing #547 as test is a good idea, I would try that!

eknoes:
I would add a new API Controller, or should I implement a new function into the files API for hiding an Album?
`/api/files/hide?album={path}`
And this should then be called after clicking the button, and should create a .nomedia file in the given path. Or am I totally wrong?

oparoz:
My idea was more to just upload a file call `.nomedia` . The Javascript should already have everything you need to upload a file, but instead of opening the dialogue to let the user pick a file, just send the filename.

Makes sense?

oparoz:
This is how to add a file creating entry:
https://github.com/owncloud/files_texteditor/blob/master/js/editor.js#L604-L642

We can see that to create the text file, it calls a few filelist specific methods, especially `fileList.createFile(name)`

That's the method's code: 
https://github.com/owncloud/core/blob/a7e7f5e18098ff0f536e58a7e1e5dfdc825e9086/apps/files/js/filelist.js#L1949-L1999

That you could try to recreate locally, in this file:
https://github.com/owncloud/gallery/blob/master/js/upload-helper.js
- We already have `isFileNameValid`
- The DAV client is available from `Gallery.filesClient`
- `getCurrentDirectory` is `Gallery.currentAlbum`
- `getUniqueName` can be written like this: https://github.com/owncloud/gallery/blob/master/js/upload-helper.js#L27-L29
- `_getWebdavProperties` might be a problem, or be as simple of using https://github.com/owncloud/gallery/blob/master/js/upload-helper.js#L25-L42

eknoes:
Oh, thanks for all the help. I will implement it tomorrow, the way you suggest sounds good!

eknoes:
Now I am a little bit stuck on the "What should happen after clicking the button"-question. I would propose, that the gallery app should go one album up, e.g. if you are on `#Pictures/Album1` there is just a `window.location="#Pictures` executed.

I just removed the getUniqueName, because it is imho unneccesary for our use case, if the file already exists the `filesClient.putFileContents` will take care of those errors.

oparoz:
> Now I am a little bit stuck on the "What should happen after clicking the button"-question. I would propose, that the gallery app should go one album up, e.g. if you are on #Pictures/Album1 there is just a window.location="#Pictures executed.

Well, the idea is to stay on the same album, so that the user can immediately see that's it's been hidden. If you go up one album and have dozens of albums, then you have to check carefully to make sure the one you wanted to hide is really gone.
You agree?

> I just removed the getUniqueName, because it is imho unneccesary for our use case, if the file already exists the filesClient.putFileContents will take care of those errors.

Well, the goal of that method is to avoid having to make a HTTP request to get the answer, but in our case, we don't need it because we wouldn't be able to be in the album if it was blacklisted.
So, all good.

eknoes:
> Well, the idea is to stay on the same album, so that the user can immediately see that's it's been hidden. If you go up one album and have dozens of albums, then you have to check carefully to make sure the one you wanted to hide is really gone.
> You agree?

All right!

> Well, the goal of that method is to avoid having to make a HTTP request to get the answer, but in our case, we don't need it because we wouldn't be able to be in the album if it was blacklisted.
> So, all good.

Okay. But I implemented the method as it was, so it is possible to create any file. Should it be just for this one usecase?

oparoz:
> Okay. But I implemented the method as it was, so it is possible to create any file. Should it be just for this one usecase?

Up to you. We might need it when adding a "create folder" entry, if ever, but that can be looked at when working on that issue.

eknoes:
So everything works now. How should I suggest extending the wiki, or don't we need to explain how to add a custom submenu to the [+] button?

oparoz:
@eknoes - It's a good idea to write something about extending the menu. I don't think it exists in the core documentation yet.

Create/edit `gallery.cnf` would be useful, but I don't think launching files_editor would work as-is
Are there any other use cases besides creating albums?

eknoes:
I think creating and editing gallery.cnf is a quite nice idea when files_texteditor is enabled, but I would do this in a seperate pull Request.

We coult just take the docs of [https://doc.owncloud.org/](https://doc.owncloud.org/server/9.0/developer_manual/app/js.html#extending-the-new-menu-in-the-files-app) for extending the menu, but we have to remove the lines with 

```
templateName: 'templateName.ext',
fileType: 'file',
```

So it would be:

# Extending the [+] menu in the gallery app

```
var myFileMenuPlugin = {
    attach: function (menu) {
        menu.addMenuEntry({
            id: 'abc',
            displayName: 'Menu display name',
            iconClass: 'icon-filetype-text',
            actionHandler: function () {
                console.log('do something here');
            }
        });
    }
};
OC.Plugins.register('OCA.Gallery.NewFileMenu', myFileMenuPlugin);
```

This will register a new menu entry in the [+] menu of the gallery app. The method attach() is called once the menu is built. This usually happens right after the click on the button.

eknoes:
I just squashed my commits. Can you give me a hint how to write a test for that? Or is it not necessary?

oparoz:
> I think creating and editing gallery.cnf is a quite nice idea when files_texteditor is enabled, but I would do this in a seperate pull Request.

Definitely in another PR :) 

> Can you give me a hint how to write a test for that? Or is it not necessary?

Well, the problem is that there is no JS test structure in place. It never existed. I wrote all the tests for PHP and would love to start using Jasmine, sinon, etc. like in `core`, but never took the time to set it up...
If you have experience with JS tests and want to give it a go, be my guest, otherwise, it's not yet a requirement.

oparoz:
We do have an acceptance test suite which uses PhantomJS locally and Firefox on Travis though. Those tests are quite slow to run, but pretty easy to write as you just simulate what a user would do.

eknoes:
Unfortunately I don't have any experience in writing Tests, so I am also not into JS tests. In the next few days I will try PhantomJS.

oparoz:
OK :)

js/upload-helper.js
```diff
@@ -69,6 +69,72 @@ var FileList = {
 		});
 	},
 
+
+	/**
+	 * Create an empty file inside the current album.
```
oparoz, line 74: ~~"Create a `.nomedia`  file" instead maybe? And explaining that this hides the current album.~~

Never mind, even if we use an editor, we'll still need to create an empty file first.

oparoz:
I've finally had time to test and I think it works great :)

It also allowed me to find a couple of unrelated bugs :)

Regarding `js/galleryhidealbum.js`, maybe you should rename it to something more generic regarding newfilemenu plugins and we could add all future plugins in there?

oparoz:
Added your page to the wiki: https://github.com/owncloud/gallery/wiki/Extending-the-%5B-%5D-menu

oparoz:
@eknoes - Let me know when you have some time to finish this. I've noticed that you've tried to rename the file, but there are now 2 versions. Github messes things up sometimes..
Also it should be `newfilemenuplugins.js` and not `newfilemenuhideplugin.js`

eknoes:
I messed up pretty much, sorry. You can see the "better" Pull Request here:  #643 @oparoz

PR #628, opened 2016-03-26
The ending slash is optional

oparoz:
Fixes regression in #620.

URLs in files apps can end with or without a `/` 

@rullzer @PVince81


oparoz:
This goes back to square one as it loads the scriot in any files_abc app.

PR #646, opened 2016-05-01
remove the iframe transport reference which no longer exists in owncl…

Spacefish:
Gallery seems to reference an jquery.iframe-transport for the files app, which no longer exists in owncloud 9.1

### DIFF
--- a/templates/part.content.php
+++ b/templates/part.content.php
@@ -40,7 +40,6 @@ script(
 		'upload',
 		'file-upload',
 		'jquery.fileupload',
-		'jquery.iframe-transport'
 	]
 );
 style(


oparoz:
Thanks @Spacefish . That was probably for <IE11.

@imjalpreet could you do a quick check to make sure uploading still works as expected on IE11 and Edge?

imjalpreet:
@oparoz Actually I work on Ubuntu 14.04. I am sorry but I am not quite sure, whether I can use IE11 and Edge on Ubuntu? Can you please tell me if there is a way to do so? or do I have to setup the development environment in Windows as well?

oparoz:
@imjalpreet - Just use the official VMs from Microsoft: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/windows/

imjalpreet:
@oparoz Cool! Thanks! :+1:

Issue #649, opened 2016-05-12
Image removing straight from the slideshow or the gallery (delete button).


tempico:
## Feature request

**User type**: Logged-in user
### Description

Image removing straight from the slideshow or the gallery (delete button).
### Benefit / value

In case of 1000 party pics in one folder it will save time to remove unneeded stuff from gallery without switching to "files" view. In future "remove" functionality can be upgraded to "removing with multiple file selection" functionality.
### Risk / caveats

If pictures will be (re)moved to trash folder I do not expect any risk here.
### Sponsorship

I can donate $20-50, but this depends.

oparoz:
It's a bit more complicated than I initially thought because we need to refresh the photowall as well...

## Design
- Add bin icon

## Code

On delete:
- move to bin
- load next slide
- remove image from slideshow array
- remove image from photowall cache
- set a flag which tells the photowall to refresh when slideshow ends

tempico:
Maybe it is possible to take "files" functionality and just use bigger preview pictures? "Files" have selection/removing/sharing/moving/"file_wall_update_on_file_remove" functionality already.

oparoz:
Unfortunately, it's not that simple, the photowall is built differently and has a different use case. I think at some point I'll add a select mode to the photowall allowing for easy selection for move and delete, but the first step will be to have a big bin onto which thumbnails can be dropped.
There is also a request to add the 3-dots menu on each thumbnail which should help.

tempico:
> move and delete

and maybe "download as zip/tar" too? I saw good implementation of this function in AirDroid gallery.

manishbisht:
I would like to work on this issue.
So when the delete function is called the following things will be done :
-Image will be moved to bin (to move image in the same bin which is used in the files)
-Load next slide
-Remove image from slideshow array
-Remove image from photowall cache

About the another addition "download as zip/tar"
When the multiple images are selected in the gallery a button showing "download as zip" can be shown near the sorting buttons.

oparoz:
You can go ahead @manishbisht . Just focus on the delete button since multi-select is not available yet.

tempico:
@manishbisht good, thank you!
What about single picture download button (like in gallery+)? Is it possible to implement this functionality too?

oparoz:
> What about single picture download button (like in gallery+)? Is it possible to implement this functionality too?

That's already implemented.

tempico:
@oparoz ,
[...]

PR #655, opened 2016-05-19
Added Delete Button in Slideshow

manishbisht:
Fixes: #649 

Licence: AGPL
### Reviewers

@oparoz @tempico


oparoz:
@manishbisht - Still the same problem. The easiest thing to do would be to create a new branch from master and to cherry-pick your commit.

Issue #663, opened 2016-06-06
Update test setup to use Share 2.0 to work around a bug in core


oparoz:
Tests broke back in April due to a bug in core. It will eventually be fixed, but migrating to Share 2.0 as a workaround shouldn't be too difficult and will have to be done eventually.

https://github.com/owncloud/core/issues/24981

oparoz:
There is another problem with legacy shares: https://github.com/owncloud/core/issues/25065

PR #665, opened 2016-06-09
Do not allow access to write-only shares

LukasReschke:
Adds checks as required for https://github.com/nextcloud/server/pull/19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,7 @@ before_script:
 
 script:
   # Test lint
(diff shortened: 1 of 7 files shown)


LukasReschke:
@oparoz As discussed.

LukasReschke:
@oparoz I don't quite understand why the tests are failing here. Could you help me on that? That would be awesome 🚀

oparoz:
Because core is broken: https://github.com/owncloud/gallery/issues/663, so until either core is fixed or a workaround is implemented, all tests will fail on master.

The backport is not affected, there is only 1 failed test which you can probably easily fix.

oparoz:
Closed in favour of #672

PR #669, opened 2016-06-12
Tokens are too loose

oparoz:


oparoz:
All tests pass locally

PR #672, opened 2016-06-12
Block write-only shares

oparoz:
Blocks access to write-only shares such as those created with the enterprise app `files_drop`

### DIFF
--- a/middleware/envcheckmiddleware.php
+++ b/middleware/envcheckmiddleware.php
@@ -16,6 +16,7 @@
 
 namespace OCA\Gallery\Middleware;
 
+use OCP\Constants;
 use OCP\IRequest;
 use OCP\IURLGenerator;
 use OCP\ISession;
@@ -162,6 +163,7 @@ class EnvCheckMiddleware extends CheckMiddleware {
 
 		$this->checkShareIsValid($share, $token);
 		$this->checkItemType($share);
+		$this->checkShareIsNotWriteOnly($share);
 
 		return $share;
 	}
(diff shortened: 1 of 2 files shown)


oparoz:
@LukasReschke - This replaces your PR #665

oparoz:
@LukasReschke - I did not add the log bypass commit (yet). Is the log getting spammed without it?

oparoz:
All tests pass locally

LukasReschke:
> Is the log getting spammed without it?

Yep. Every access to such a share in the normal upload view triggers a log entry otherwise. To reproduce:
1. Create a folder
2. Share it via link
3. Make it editable
4. Check "Hide file listing".
5. Access the link

The error log will be spammed :(

oparoz:
But this only happens if someone tries to open Gallery using a token with write-access, no? That's one log entry per mistake. When people upload using the files_drop app, then nothing is getting logged in Gallery.
Am I missing something?

LukasReschke:
> But this only happens if someone tries to open Gallery using a token with write-access, no? That's one log entry per mistake. When people upload using the files_drop app, then nothing is getting logged in Gallery.
> Am I missing something?

The files_drop app as implemented in Nextcloud is not a dedicated app but simply a sharing option. So you access the share via `index.php/s/{token}` and from there you just get an upload interface.

The problem there is that the Gallery app is also injecting itself into that view via JS. So the JS gets loaded and in there an exception is thrown that is then logged.

oparoz:
OK, got it. Exceptions were added for `files_drop`, but if the URL is the same... 

Would be cool to use ["loadAdditionalScripts"](https://github.com/owncloud/core/issues/22328) and disable it when the share is write-only.

oparoz:
Closing as this will not be added to Gallery for ownCloud

PR #658, opened 2016-05-23
Added Delete Button in Slideshow

manishbisht:
Fixes: #649 

Licence: AGPL
### Tested on
- [x] Google Chrome
### Reviewers

@oparoz @tempico

PR #676, opened 2016-06-17
Added Delete Button in Slideshow

oparoz:
Local duplicate of #658 and rebased.

Fixes: #649
- [x] Disable on read-only shares
- [x] Disable on images which can be deleted
- [x] Fix icon
- [x] Fix current image index for case where image is the last one
- [x] Fix delete from Files
- [ ] Fix wrong method for webdav endpoint


oparoz:
There is a bug when going in reverse direction.

manishbisht:
Bug Resolved @oparoz :)

oparoz:
Thanks, I'm guessing you removed that extra delete statement.

You can find my fix and cleanup here:
https://github.com/interfasys/galleryplus/pull/47/commits/96dffb72d895e30a6d3bf52ee31c232ff4b0db52

manishbisht:
Yes, I have only removed the extra delete statement @oparoz  :)

oparoz:
OK, I've pushed my cleanup as there were other issues.
There is still one problem left if you want to take a look at it: The icon should not be shown on shared albums unless the share is read-write.

manishbisht:
Ok. I will look into this after my exams. Can we close the original issue #649 @oparoz

oparoz:
Done

oparoz:
Actually, we can't close the issue as it isn't merged.

manishbisht:
Can we merge this PR with owncloud gallery ?

oparoz:
Not before there is a stable branch to merge to.

oparoz:
And before the bugs are solved, of course.

manishbisht:
Ok. Actually I want to claim the bounty for this issue :p

oparoz:
> Actually I want to claim the bounty for this issue :p

Then you need to fix the bugs.

manishbisht:
Ok. I will try to fix this bug @oparoz :)

oparoz:
OK, don't forget to pull the changes and then to create a new commit containing ONLY your changes.

manishbisht:
I think the process to check the share properties during files deletion must be present in the files app also.

oparoz:
I assume so since a read-write share can be fully managed from the Files side. 
That's the `files_sharing` app btw.

oparoz:
And there is another issue. The delete icon used doesn't have any shadows and thus make it invisible on a light background.

hidden_delete_icon

oparoz:
OK, so now the only thing missing is the icon.

manishbisht:
@oparoz I have shared one album and opened it in browser. It is showing me Gallery.token is undefined and when I opened the same image in slideshow it shows Gallery.token.is undefined

manishbisht:
Its working with latest commit :+1:

oparoz:
2 more problems:
- ~~Fix delete from Files~~
- Fix wrong method for webdav endpoint

PR #680, opened 2016-06-19
Added delete button

manishbisht:
Fixes: #649 
- [x] Disable on read-only shares
- [x] Disable on images which can be deleted
- [x] Fix icon
- [x] Fix current image index for case where image is the last one
- [x] Fix delete from Files
- [x] Fix wrong method for webdav endpoint

Licence: AGPL
### Tested on
- [x] Ubuntu/Google Chrome
### Reviewers

@oparoz @tempico


tempico:
I found small problem.
For example, I have folder with 3mb pictures each. When I remove one using delete button, browser sends delete request via webdav to server. After server's answer, browser loads another 3mb picture. All process can take up to 10 seconds and all this time I can see only black screen. Maybe we can add some spinner?

oparoz:
@manishbisht - It's very difficult to understand what the changes are from your commits, but I assumed you simply added this?

```
if (OC.PERMISSION_DELETE === 0) {
    return;
}
```

How did you test?

manishbisht:
@oparoz I have found this in the files app delete function. You can cross verify from it.

oparoz:
@manishbisht - Finding something and testing it are not the same thing. I don't get `0` on a shared album, so I doubt the button hides properly, but you've indicated that you've tested on `Ubuntu/Google Chrome`, so how did you test?

oparoz:
@tempico - I'm testing on Gallery+ and see a spinner when the next image is loading, so maybe try with a zip from stable9 at [`interfasys/galleryplus`](https://github.com/interfasys/galleryplus/tree/stable9) ?

manishbisht:
```
// if there is no permission to delete do nothing
    if((context.$file.data('permissions') & OC.PERMISSION_DELETE) === 0) {
        return;
    }
```

@oparoz I have found this code that the files app uses to check that we have the permissions to delete. I have not tested it for shared images but I have checked it for normal images OC.PERMISSION_DELTE returns 8

manishbisht:
@tempico I am also seeing the spinner using ownCloud 9.1.0 beta 2 (git)

oparoz:
> I have found this code that the files app uses to check that we have the permissions to delete

And you've removed half of the condition.

Luckily, it shouldn't be too hard to add the 2nd half since all albums have their permissions assigned to them in the JS.

> I have not tested it for shared images but I have checked it for normal images OC.PERMISSION_DELTE returns 8

If you had tested it, you would have seen that, it also returns 8 on a read-only share. It's the other half of the check which makes the difference.

manishbisht:
@ I have added
`if(Gallery.imageMap[imagePath].sharedWithUser === false)`
So that the deletion is only possible when shareWithUser === false By this only the original user will only be able to delete the image.

oparoz:
That won't work as this value indicates that the image is shared, not that it's currently being viewed in a shared album.

oparoz:
I think you were on the right path when you looked at permissions, but you just need to use what's available in Gallery or add additional parameters to the template. The 1st option seem easier.

manishbisht:
Thanks @oparoz for your suggestion. I have changed the above line with `if ((Gallery.albumMap[Gallery.currentAlbum].permissions & OC.PERMISSION_DELETE) === 0) {` If this condition is matched we will not delete the image. I think it will work now. :)

Also how can I test the shared feature ?

manishbisht:
@oparoz Is there any way to show notifications to the user that "you don't have enough permissions".

oparoz:
> Also how can I test the shared feature ?
1. Share an album via link
2. Go to the link
3. Click on an image

oparoz:
> Is there any way to show notifications to the user that "you don't have enough permissions".

There is.
A message is shown to the user if he tries to move an image in a read-only folder. A better option would be to disable moving images, just like uploading is disabled.
Showing a notification doesn't work with delete though, we don't want to show actions which are not possible.

oparoz:
I've just tested `(Gallery.albumMap[Gallery.currentAlbum].permissions & OC.PERMISSION_DELETE)` from the console and I get 8 both times, so I checked `fileactions` where you got the idea from and here is the list of problems:
- The test has to be put elsewhere so that we can control if we show the icon or not. It's bad practice to show the icon and then tell the user he can't do that.
- The test you're using works on a file, so adapt your modification to detect if a file can be deleted, even when in "normal" mode
- Prepare a test setup with various scenarios so that we can all test and validate the solution

manishbisht:
@oparoz To implement the shadow to the delete button I think we have to add a new image in the core that contains the shadow or we can do it with css. Which will be more better ?

oparoz:
> To implement the shadow to the delete button I think we have to add a new image in the core that contains the shadow or we can do it with css

You can do it via CSS, like on stable8.2

Also regarding your approach. I'm assuming that `fileactions` is what is being used by `files_sharing`. If that's not the case, then it won't work.

manishbisht:
@oparoz I have opened the image after sharing the album in both editing mode and non-editing mode. It opened on the slideshow but on the slideshow download button is hidden always. Can we do the same thing for the delete button ?

oparoz:
> Can we do the same thing for the delete button  ?

Unfortunately, no, because that's just a usability thing. You can already download from the filelist, so I found it to be redundant, but I think I'm going to remove that behaviour as some people prefer to be able to download an image from the slideshow.

The delete action should not be offered if you can't perform that action.

oparoz:
I've just checked the permissions and they're wrong on the public side. They should switch to 17, but are still 27.

oparoz:
OK, so for now, I've disabled the delete button on public shares. There is no way to delete content on a public share, a user can only read and/or upload.

oparoz:
The remaining step in terms of logic is to disable the button on normal shares for images which don't have the delete permission

manishbisht:
@oparoz For implementing the shadow I have used the css which have done in stable 8.2

```
#slideshow input.icon-view-download {
    background-image: url("../img/download.svg");
    background-size: 26px 26px;
    -webkit-background-size: 26px 26px;
    background-color: #fff;
    height: 38px;
    width: 38px;
}
```

But when I applied the same thing for ".deleteImage" button. It doesn't make any effect.

oparoz:
> For implementing the shadow I have used the css which have done in stable 8.2

There is no shadow in the CSS code you posted. What I wanted you to see was the use of `background-image`

Now you need to modify the SVG image to add a shadow, then generate a PNG and finally optimize both files.

manishbisht:
@tempico @oparoz I have added the bin icon and fixed all the bugs (Doesn't added PNG as IE < 11 no more supported by OC 9+). I was busy in my semester exams will be contributing regularly in next 2-3 days :)

manishbisht:
Any Update on the PR @oparoz

oparoz:
Thanks for the icon. If you could add it as a separate commit, I could then add it to the PR that will be merged.

You also need to make a slight change to the icon, you have a standard deviation of 4 and I think it would work best with 6.

manishbisht:
@oparoz I have seperated the commit :)

oparoz:
Thanks!

oparoz:
#676 has been merged, including your original commit and your icon. Thank you for your contribution!

PR #652, opened 2016-05-15
Fix of issue #23066

TDannhauer:
Fixes Bug owncloud/core#23066

PR #682, opened 2016-06-28
[stable9] Fix of issue #23066

butonic:
backport of https://github.com/owncloud/gallery/pull/652 to fix https://github.com/owncloud/core/issues/23066 in stable9

@oparoz @PVince81


PVince81:
Makes sense :+1:

oparoz:
The session tests are failing, so either they need to be altered or the fix is not as simple as casting to string.

oparoz:
Ok, so it's simply a matter of fixing the tests and casting another ID: https://github.com/interfasys/galleryplus/pull/52/files

PVince81:
Added extra cast in `checkPassword` and adjusted unit tests.

PR #684, opened 2016-07-03
Add more escaping

oparoz:
@LukasReschke

PR #688, opened 2016-07-06
[stable8.2] Add more escaping

PVince81:
Backport of https://github.com/owncloud/gallery/pull/684 to stable8.2.
Note: there were conflicts.

@oparoz @DeepDiver1975

### DIFF
--- a/js/vendor/owncloud/share.js
+++ b/js/vendor/owncloud/share.js
@@ -296,7 +296,7 @@
 		showDropDown:function(itemType, itemSource, appendTo, link, possiblePermissions, filename) {
 			var data = OC.Share.loadItem(itemType, itemSource);
 			var dropDownEl;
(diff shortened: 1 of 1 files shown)


oparoz:
> Note: there were conflicts.

Did you get JS errors?

PVince81:
I didn't. Also I realized later on that itemSource is an integer here so this PR might not really be needed.

oparoz:
> Also I realized later on that itemSource is an integer here so this PR might not really be needed.

You're thinking of a different PR ;)

PVince81:
I redid a quick test with a debugger and evaluated itemSource: it's an integer.

oparoz:
What about itemType?

PVince81:
Hmm, I thought itemType is only limited to known values like "file", "folder".
Fair enough, let's go the safe route.

Reopened.

PR #689, opened 2016-07-07
Fix for string id

PVince81:
Gallery part for https://github.com/owncloud/core/issues/23066
- stable9.1: https://github.com/owncloud/gallery/pull/690
- stable9: https://github.com/owncloud/gallery/pull/682

@butonic @oparoz 

will have a look at the unit tests that were mentionned to fail

### DIFF
--- a/middleware/envcheckmiddleware.php
+++ b/middleware/envcheckmiddleware.php
@@ -267,7 +267,7 @@ class EnvCheckMiddleware extends CheckMiddleware {
 		$newHash = '';
 		if ($this->shareManager->checkPassword($share, $password)) {
 			// Save item id in session for future requests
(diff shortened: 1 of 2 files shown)


oparoz:
Is this really necessary for 9.1+ ? I thought it only was a problem on 9.

PVince81:
I don't have an env to reproduce the original issue so better be sure.
I'm not aware of any code in 9.1 that would make this fix unnecessary.

oparoz:
OK, thanks. I never had any issue on 9.1, but as you said, it doesn't hurt to keep this change in.

PVince81:
I'll prepare the backports.
The one from @butonic is incomplete, will finish it.

PR #691, opened 2016-07-12
Use variable name instead of hard-coded default

LukasReschke:
cc @oparoz @MorrisJobke @blizzz

### DIFF
--- a/templates/public.php
+++ b/templates/public.php
@@ -76,11 +76,11 @@ style(
 						  data-owner="<?php p($_['displayName']) ?>"
 						  data-name="<?php p($_['filename']) ?>">
 									<button id="save-button"><?php p(
-											$l->t('Add to your ownCloud')
+											$l->t('Add to your %s', [(new \OCP\Defaults())->getName()])
 										) ?></button>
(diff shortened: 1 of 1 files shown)


blizzz:
LGTM

templates/public.php
```diff
@@ -76,11 +76,11 @@
 						  data-owner="<?php p($_['displayName']) ?>"
 						  data-name="<?php p($_['filename']) ?>">
 									<button id="save-button"><?php p(
-											$l->t('Add to your ownCloud')
+											$l->t('Add to your %s', [(new \OCP\Defaults())->getName()])
 										) ?></button>
 									<form class="save-form hidden" action="#">
 										<input type="text" id="remote_address"
-											   placeholder="example.com/owncloud"/>
+											   placeholder="cloud.example.com"/>
```
jancborchardt, line 83: Same as in https://github.com/owncloud/gallery/pull/692, a better address would be `example.com/cloud`. (I’d just quickly change it myself but don't have push access anymore. ;)

oparoz:
I would prefer it if that variable was sent from the controller, so FYI this may not be merged as-is

PR #692, opened 2016-07-12
[stable9] Use variable name instead of hard-coded default

LukasReschke:
Backport of https://github.com/owncloud/gallery/pull/691

cc @oparoz @MorrisJobke @blizzz

### DIFF
--- a/templates/public.php
+++ b/templates/public.php
@@ -76,11 +76,11 @@ style(
 						  data-owner="<?php p($_['displayName']) ?>"
 						  data-name="<?php p($_['filename']) ?>">
 									<button id="save-button"><?php p(
-											$l->t('Add to your ownCloud')
+											$l->t('Add to your %s', [(new \OCP\Defaults())->getName()])
 										) ?></button>
 									<form class="save-form hidden" action="#">
 										<input type="text" id="remote_address"
-											   placeholder="example.com/owncloud"/>
+											   placeholder="cloud.example.com"/>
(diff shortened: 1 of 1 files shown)


templates/public.php
```diff
@@ -76,11 +76,11 @@
 						  data-owner="<?php p($_['displayName']) ?>"
 						  data-name="<?php p($_['filename']) ?>">
 									<button id="save-button"><?php p(
-											$l->t('Add to your ownCloud')
+											$l->t('Add to your %s', [(new \OCP\Defaults())->getName()])
 										) ?></button>
 									<form class="save-form hidden" action="#">
 										<input type="text" id="remote_address"
-											   placeholder="example.com/owncloud"/>
+											   placeholder="cloud.example.com"/>
```
jancborchardt, line 83: A better placeholder would be `example.com/cloud` as that’s a more common pattern for addresses. Otherwise this looks fine.

LukasReschke, line 83: > A better placeholder would be example.com/cloud as that’s a more common pattern for addresses. Otherwise this looks fine.

I don't really like the idea of advertising using a subfolder for your Nextcloud/ownCloud 🙈

oparoz, line 83: I agree with @LukasReschke . I think it's bad practice to use a subfolder, especially if other apps are installed in other folders.

jancborchardt, line 83: Ok, good point I guess.

PR #700, opened 2016-08-01
Update info.xml

SergioBertolinSG:


ChristophWurst:
Why did you set the min-version to 9.2?

SergioBertolinSG:
To be consistent with the rest of apps.

ChristophWurst:
> To be consistent with the rest of apps.

I don't get it. Before the app targeted two major releases (9.0 and 9.1) and now it's only 9.2. Is there any reasoning for that?

ChristophWurst:
Is it because the app is shipped?

DeepDiver1975:
> Is it because the app is shipped?

basically yes - a shipped app is only used for that specific major version.

PR #721, opened 2017-05-05
Appstore makefile

IljaN:
Makefile to create a package for marketplace.

Removed npm/composer targets as all dependecies are checked in vcs.


PVince81:
Is it intended that the appstore bundle contains a "vendor" folder ?

From what I seem to remember the vendor folder is only used for running the test suites and might not be needed for a release build.

IljaN:
@PVince81 done

PR #784, opened 2018-07-31
Reformat .php_cs.dist to be similar to core

phil-davis:
Fixes #785 

This change reformats the ``.php_cs.dist`` to build its finder/exclude list in the same way that is done in ``core``, and it works.

No idea why the old way stopped working, I guess something happened in ``php-cs-fixer``  in the last few weeks.

PR #786, opened 2018-07-31
Trigger CI

phil-davis:
Demonstrate current CI results.


phil-davis:
The problem is illustrated in Travis job https://travis-ci.org/owncloud/gallery/jobs/410286745

It is fixed by #784

PR #794, opened 2018-08-28
Implement working .php_cs.dist

phil-davis:
For some reason the "old" way of doing ``exclude()`` in ``.php_cs.dist`` has stopped working.
In CI, ``php-cs-fixer`` is checking ``c3.php``
https://travis-ci.org/owncloud/gallery/jobs/421365878
It should be ignoring it.

1)  I have updated the code in ``.php_cs.dist`` to match the way it is now done in other app repos.

2) Then it fails on PHP 7.2 with outdated versions of dev dependencies, so ``composer update`` has been done.

3) Then ``php-cs-fixer`` reports new code format issues that it now finds. So those have been applied.

And finally CI passes again.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,13 +1,12 @@
 <?php
 
-$finder = PhpCsFixer\Finder::create()
+$config = new OC\CodingStandard\Config();
+
+$config
+    ->setUsingCache(true)
+	->getFinder()
 	->exclude('l10n')
 	->exclude('c3.php')
 	->in(__DIR__);
-
-$config = new OC\CodingStandard\Config();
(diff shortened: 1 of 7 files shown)


phil-davis:
Note: only the PHP 7.2 CI jobs were failing. The "old" way seems to be working still on PHP 5.6 7.0. and 7.1

phil-davis:
@patrickjahns I am just getting CI working again here - not attempting to cleanup the way the whole thing works, or to move it to drone. That is another big job.

patrickjahns (approved):
🌞  - happy CI - happy QA-Team 🌞

PR #795, opened 2018-08-28
Composer update 20180828

phil-davis:


phil-davis:
Just  the ``composer update`` and applying php-cs-fixer changes is not enough. It still gets the problem that it reports on ``c3.php``
So all the commits in #794 are needed.

PR #818, opened 2019-01-10
Trigger CI

phil-davis:
I does not seem to be running in nightly drone jobs


phil-davis:
all good - this repo still runs lint checks with ``parallel-lint`` but installs ``parallel-lint`` itself, so it works
(``parallel-lint`` was removed from core yesterday)

PR #824, opened 2019-03-13
Adjust composer files and app core min-version for 'drop PHP 5.6'

phil-davis:
- [x] adjust local `composer` files to PHP 7.0.8
- [x] bump app core min-version to `10.2`

because the app is now being tested only against core `stable10` upcoming `10.2` that has dropped PHP 5.6 support.


individual-it (approved):
an other one with a lot of stuff deleted from composer.lock I guess a similar case like https://github.com/owncloud/activity/pull/718#pullrequestreview-213951709

phil-davis:
@individual-it yes, a lot of crud waiting for someone to type `composer update`

